PR #1, opened 2018-12-15
Add currencyconverterapi.com plugin for IMP and IRR support only

paullinator:


src/currencyconverterapi.js
```diff
@@ -0,0 +1,55 @@
...
+          { fromCurrency: 'iso:USD', toCurrency: 'iso:IRR' }
+        ]
+        const isoCodesWanted = []
+        for (const pair of pairsHint) {
+          checkAndPush(pair.fromCurrency, isoCodesWanted)
+          checkAndPush(pair.toCurrency, isoCodesWanted)
+        }
+
+        const pairs = []
+        for (const isoCode of isoCodesWanted) {
```
swansontec: We could probably refactor this into `Promise.all(isoCodesWanted.map(...))` for parallelism, but this seems fine for now, especially since we have a short list.

PR #4, opened 2019-02-02
Add HERC's exchange

kylanhurt:
Same code as HERC's pull request but added export so that it is readable by the GUI. Also for simplicity.


kylanhurt:
Merging since it would need a publish to affect GUI anyway

PR #9, opened 2019-03-09
Create Totle plugin.

passabilities:


src/plugins/totle.js
```diff
@@ -0,0 +1,225 @@
...
+          gasPrice: quoteInfo.gas.price
+        }
+      }
+      io.console.info('Totle spendInfo', spendInfo)
+      const tx = await request.fromWallet.makeSpend(spendInfo)
+      tx.otherParams.payinAddress = spendInfo.spendTargets[0].publicAddress
+      tx.otherParams.uniqueIdentifier =
+        spendInfo.spendTargets[0].otherParams.uniqueIdentifier
+
+      return makeSwapPluginQuote(
```
swansontec: Since this plugin needs to send not one, but two transactions, I would just extract the contents of `makeSwapPluginQuote` and do that work inline here.

src/plugins/totle.js
```diff
@@ -0,0 +1,225 @@
...
+              pluginName: swapInfo.pluginName,
+              uniqueIdentifier: quoteInfo.id,
+              data: quoteInfo.payload.data
+            }
+          }
+        ],
+        networkFeeOption: 'custom',
+        customNetworkFee: {
+          gasLimit: quoteInfo.gas.limit,
+          gasPrice: quoteInfo.gas.price
```
swansontec: The Edge Ethereum plugin unfortunately takes the gas price in GWei. This was a mistake, but I'm not sure whether it would be easier to fix the Ethereum plugin + GUI, or just have Totle divide the number by a billion.

src/plugins/totle.js
passabilities: 👍 I actually already made this change and testing it out rn. Will hopefully push this change today.

src/plugins/totle.js
```diff
@@ -0,0 +1,328 @@
...
+  type EdgeCurrencyWallet,
+  type EdgeSwapPlugin,
+  type EdgeSwapPluginQuote,
+  type EdgeSwapRequest,
+  SwapCurrencyError
+} from 'edge-core-js/types'
+import Web3 from 'web3'
+
+import { getFetchJson } from '../react-native-io.js'
+import type { EdgeTransaction } from 'edge-core-js/src/types/types'
```
swansontec: Merge this with the import statement above.

src/plugins/totle.js
```diff
@@ -0,0 +1,328 @@
...
+}
+
+const ETH_ADDRESS = '0x0000000000000000000000000000000000000000'
+const ERC20_ABI = [ { constant: false, inputs: [ { name: '_spender', type: 'address' }, { name: '_value', type: 'uint256' } ], name: 'approve', outputs: [ { name: '', type: 'bool' } ], payable: false, type: 'function' }, { constant: true, inputs: [ { name: '_owner', type: 'address' }, { name: '_spender', type: 'address' } ], name: 'allowance', outputs: [ { name: '', type: 'uint256' } ], payable: false, type: 'function' } ]
+const totleTransferProxyAddress = '0x74758AcFcE059f503a7E6B0fC2c8737600f9F2c4'
+const provider = new Web3.providers.WebsocketProvider('wss://node.totlesystem.com')
+const web3 = new Web3(provider)
+const { toBN } = web3.utils
+
+async function getAddress (
```
swansontec: Ethereum only has one address format, so we probably don't need this.

PR #10, opened 2019-04-03
Migrate CoinCap to new API

kylanhurt:
The purpose of this task is to migrate the CoinCap plugin to the new API (/v2/assets instead of /front).

### DIFF
--- a/src/plugins/coincap.js
+++ b/src/plugins/coincap.js
@@ -14,20 +14,20 @@ export function makeCoincapPlugin (opts: EdgeCorePluginOptions): EdgeRatePlugin 
     },
 
     async fetchRates (pairsHint) {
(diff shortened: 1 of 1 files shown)


src/plugins/coincap.js
```diff
@@ -14,20 +14,20 @@ export function makeCoincapPlugin (opts: EdgeCorePluginOptions): EdgeRatePlugin
     },
 
     async fetchRates (pairsHint) {
-      const reply = await io.fetch('https://coincap.io/front')
+      const reply = await io.fetch('https://api.coincap.io/v2/assets')
       const json = await reply.json()
```
swansontec: `if (json.data == null) throw new Error('No coincap price data')`

PR #11, opened 2019-04-10
Re-add old Coincap rates (legacy)

kylanhurt:
The purpose of this task is to add back the legacy API / exchange rates for Coincap so that the exchange rates that are missing from the newer API can be used within the GUI.

### DIFF
--- a/src/plugins/coincapLegacy.js
+++ b/src/plugins/coincapLegacy.js
@@ -5,6 +5,21 @@ import {
   type EdgeRatePlugin
 } from 'edge-core-js/types'
 
+const currencyCodesToInclude = [
+  'FTC',
+  'UFO',
+  'GRS',
+  'SMART',
+  'IND',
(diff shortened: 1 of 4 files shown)


src/plugins/coincapLegacy.js
```diff
@@ -0,0 +1,39 @@
...
+      const json = await reply.json()
+
+      // Grab all the pairs which are in USD:
+      const pairs = []
+      for (const entry of json) {
+        if (typeof entry.short !== 'string') continue
+        if (typeof entry.price !== 'number') continue
+        const currency = entry.short
+
+        pairs.push({
```
swansontec: The idea was that we would filter the list of currencies to just the missing ones. Otherwise, we are back to the original problem of bad rate data for things like BTC. Please coordinate with Matthew to get a list of missing coins, and only push those to the output list.

PR #12, opened 2019-04-18
Add Nomics exchange rates

kylanhurt:
The purpose of this task is to add Nomics's API to the list of available exchange rates.

Asana Task: https://app.asana.com/0/361770107085503/1118743622578989/f

### DIFF
--- a/src/plugins/nomics.js
+++ b/src/plugins/nomics.js
@@ -1,0 +1,49 @@
+// @flow
+
+import {
+  type EdgeCorePluginOptions,
+  type EdgeRatePlugin
+} from 'edge-core-js/types'
+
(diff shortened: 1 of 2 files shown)


src/plugins/nomics.js
```diff
@@ -0,0 +1,49 @@
...
+  if (apiKey == null) {
+    throw new Error('No Nomics exchange rates API key provided')
+  }
+  return {
+    rateInfo: {
+      displayName: 'Nomics'
+    },
+
+    async fetchRates (pairsHint) {
+      console.log(
```
swansontec: If you use `io.console.log` instead of `console.log`, it will go over the bridge to the main app. Not a dealbreaker, since we don't necessarily need these logs.

PR #15, opened 2019-05-02
CoinZark Swap Plugin

gertjaap:
This pull request includes the changes for the integration of [CoinZark](https://coinzark.com/) as an Exchange provider in the Edge wallet.  See also [The related PR in `edge-react-gui`](https://github.com/EdgeApp/edge-react-gui/pull/1349)

### DIFF
--- a/src/swap/coinzark.js
+++ b/src/swap/coinzark.js
@@ -1,0 +1,106 @@
+// @flow
+
+import {
+  type EdgeCorePluginOptions,
+  type EdgeSwapPlugin,
+  type EdgeSwapPluginQuote,
+  type EdgeSwapRequest,
+  SwapCurrencyError
+} from 'edge-core-js/types'
+
+import { makeSwapPluginQuote } from '../swap-helpers.js'
+
+const swapInfo = {
+  pluginName: 'coinzark',
+  displayName: 'CoinZark',
+  supportEmail: 'support@coinzark.com'
+}
+
+/* type CoinZarkQuoteJson = {
+  swap_id: string,
+  created_at: string,
+  deposit_address: string,
+  deposit_amount: number,
+  deposit_currency: string,
+  spot_price: number,
+  price: number,
+  price_locked_at: string,
+  price_locked_until: string,
+  withdrawal_amount: number,
+  withdrawal_address: string,
+  withdrawal_currency: string,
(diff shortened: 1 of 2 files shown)


src/swap/coinzark.js
```diff
@@ -0,0 +1,272 @@
...
+      // Fetch the supported currencies
+      const currencies = await get('swap/currencies')
+      let fromCorrect = false
+      let toCorrect = false
+
+      // Loop through the currencies and find the requested ones.
+      // CoinZark will return canDeposit / canReceive as status of the
+      // coins. The coin we want to exchange from should have canDeposit enabled
+      // and the coin we want to exchange to should have canReceive enabled.
+      if (!(currencies === null || currencies.result === null)) {
```
swansontec: We prefer double-equals with null (` == null`). This is the one exception to the `===` everywhere rule, since the `==` catches both null and undefined, and basically means "doesn't exist".

swansontec: Actually, reading it again, `(currencies != null && currencies.result != null)` might be even more obvious.

src/swap/coinzark.js
```diff
@@ -0,0 +1,272 @@
...
+        for (const curr of currencies.result) {
+          if (curr.id === fromCurrencyCode && curr.canDeposit === 1) {
+            fromCorrect = true
+          }
+
+          if (curr.id === toCurrencyCode && curr.canReceive === 1) {
```
swansontec: Could we treat `canReceive` like a boolean, removing the `=== 1` part? Otherwise, if the API ever returned `true` in the future, this would break.

swansontec (dismissed):
Generally looks good. I just noticed a few minor style things, which you can either fix or leave at your discretion.

swansontec:
A note on scheduling: We are currently in QA on a release right now, so master will hopefully be open for merging early next week.

src/swap/coinzark.js
gertjaap: Agreed, i've changed this.

src/swap/coinzark.js
gertjaap: The API returns a number here. If the API ever changes, it will be published as a new version and will require a change in the plug-in regardless.

PR #23, opened 2019-06-06
adjustments and fixes for faast plugin

nathanielwood:
Hey guys,

We've made some adjustments to the faast API and I've adjusted the plugin here to improve reliability. Some notes:

- I had given the wrong impression that the quotes were only estimates, as I was thought it was done a different way. With either the `deposit_amount` or `withdrawal_amount` provided, the price is guaranteed until the expiration (`price_locked_until`). So I've flipped that flag in the call to `makeSwapPluginQuote`.
- It will now use our /address endpoint to validate the addresses, and will use the `standardized` version of the address that we provide for swap creation. This is to help LTC and BCH swaps (eg. for LTC it converts the address from an '3...' to a 'M...' address.
- There's a limit check when the `quoteFor` is "from" as well as "to" now, by checking min/max that are now provided from the /price endpoint.

I'm sure I forget to mention other changes. Hope this helps though, thanks!

### DIFF
--- a/src/swap/faast.js
+++ b/src/swap/faast.js
@@ -46,6 +46,13 @@ type FaastQuoteJson = {
   terms?: string
 }
 
+type FaastAddressJson = {
+  valid: boolean,
+  blockchain: string,
+  standardized: string,
+  terms?: string
+}
+
 const dontUseLegacy = {
   DGB: true
 }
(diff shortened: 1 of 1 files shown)


nathanielwood:
A bunch on my end, just running edge-react-gui locally in Xcode with the updated build and ensuring it was working as intended. Not sure if there was any automated testing I was missing out on though.

swansontec:
Sorry for the long delay on this PR. We are running this through our QA process this week, targeting the next release.

PR #24, opened 2019-06-11
fix error handling for inactive currency

ahujsak:
fixes : https://app.asana.com/0/361770107085503/1124410096045732

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # edge-exchange-plugins
 
+# 0.6.5 (2019-06-04)
+
+- fix error when currency is temporarily disabled
+
 # 0.6.4 (2019-06-04)
 
 - fix amount string instead of number error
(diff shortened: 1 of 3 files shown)


src/swap/changelly.js
```diff
@@ -94,18 +94,17 @@ async function getAddress (
     : addressInfo.publicAddress
 }
 
-function checkReply (reply: Object, request?: EdgeSwapRequest) {
```
swansontec: I would have just made the `request` mandatory - it has these two parameters inside it, plus other things other error types might need in the future. If Changelly ever has region restrictions, you will need `request` for the `SwapPermissionError` constructor, for example.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # edge-exchange-plugins
 
+# 0.6.5 (2019-06-04)
```
swansontec: I generally like to see the release notes + version bump in a separate commit whose commit message *is* the version, like "v0.6.5".

src/swap/changelly.js
ahujsak: ok, I will make that change

swansontec:
Thanks, looks good.

PR #36, opened 2019-07-26
Integrate GoDex as swap partner

kylanhurt:


kylanhurt:
Still need to find out if `isEstimate` should be true or false.

PR #42, opened 2019-09-03
Set `isEstimate` to false and calculate guaranteed rate values.

passabilities:


src/swap/totle.js
```diff
@@ -10,6 +10,7 @@ import {
   NoAmountSpecifiedError,
   SwapCurrencyError
 } from 'edge-core-js/types'
+import BN from 'bignumber.js'
```
swansontec: We use `biggystring` for this type of bignum math. The biggystring library is a wrapper around `bn.js` (v4.11.7). So, if you could use either bn.js (small change to your code) or biggystring (bigger change to your code), that would reduce our dependency footprint.

passabilities: @swansontec I don't see either of those dependencies listed in the `package.json`. Are they imported from somewhere else or are you saying to install `bn.js`?

swansontec: Well, biggystring uses bn.js as a transitive dependency, so using bn.js directly could presumably re-use the same library. However, I am less worried about this now. If this is the only issue, we can move forward.

PR #43, opened 2019-09-20
Implement pegged rates for WETH, TBTC, and BRZ

kylanhurt:


src/rate/coinbase.js
```diff
@@ -45,8 +48,14 @@ export function makeCoinbasePlugin (
           toCurrency,
           rate
         })
+        if (DUPLICATE_RATE_MAP[toCurrency]) {
```
swansontec: This was supposed to be a *new* rate provider. It shouldn't affect any of the existing plugins in any way. The new plugin's `fetchRates` method would just return some fixed JSON data.

kylanhurt:
Superseded by: https://github.com/EdgeApp/edge-exchange-plugins/pull/44

PR #44, opened 2019-09-20
Implement fake exchange rate plugin

kylanhurt:


kylanhurt:
Already tested in GUI and works like a charm

PR #46, opened 2019-09-26
Disable USDT for Changelly (wrong parent chain)

kylanhurt:


src/swap/changelly.js
```diff
@@ -19,6 +19,8 @@ import utf8Codec from 'utf8'
 import { getFetchJson } from '../react-native-io.js'
 import { makeSwapPluginQuote } from '../swap-helpers.js'
 
+const INVALID_CURRENCY_CODES = ['USDT']
```
swansontec: An easier way to do this is to say `INVALID_CURRENCY_CODES = { USDT: true }`. Now, instead of `INVALID_CURRENCY_CODES.includes(cc)`, you can do `INVALID_CURRENCY_CODES[cc]`, which would be faster & simpler.

PR #48, opened 2019-10-02
Implement currencyCode transcription for inconsistent currency codes

kylanhurt:
The purpose of this task is to allow our swap partners to be able to exchange currencies whose Edge currencyCodes are ambiguous (because they do not identify the parent chain, eg USDT).

### DIFF
--- a/src/swap/changelly.js
+++ b/src/swap/changelly.js
@@ -23,6 +23,13 @@ const INVALID_CURRENCY_CODES = {
   USDT: true
 }
 
+// Invalid currency codes should *not* have transcribed codes
+// because currency codes with transcribed versions are NOT invalid
+const CURRENCY_CODE_TRANSCRIPTION = {
+  // Edge currencyCode: exchangeCurrencyCode
+  USDT: 'USDT20'
+}
+
 function hmacSha512 (data: Uint8Array, key: Uint8Array): Uint8Array {
   const hmac = hashjs.hmac(hashjs.sha512, key)
   return hmac.update(data).digest()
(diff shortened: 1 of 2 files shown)


src/swap/changelly.js
```diff
@@ -153,6 +160,7 @@ export function makeChangellyPlugin (
       userSettings: Object | void
     ): Promise<EdgeSwapPluginQuote> {
       if (
+        // if either currencyCode is invalid *and* doesn't have a transcription
```
swansontec: This seems like it will always throw - I don't see where we use the `CURRENCY_CODE_TRANSCRIPTION` in this check. Actually, since `CURRENCY_CODE_TRANSCRIPTION` handles this case, I think the entire check can just go away.

src/swap/changelly.js
```diff
@@ -255,7 +272,9 @@ export function makeChangellyPlugin (
       const quoteInfo: FixedQuoteInfo = sendReply.result
       const spendInfoAmount = await request.fromWallet.denominationToNative(
         quoteInfo.amountExpectedFrom,
-        quoteInfo.currencyFrom.toUpperCase()
+        // need to verify that this is okay
+        // why use currencyCode from quoteInfo in the first place?
+        request.fromCurrencyCode.toUpperCase()
```
swansontec: Yes, this looks OK. I don't know why Allen did it this way, since the request version is the official one.

src/swap/changelly.js
```diff
@@ -23,6 +23,13 @@ const INVALID_CURRENCY_CODES = {
   USDT: true
```
swansontec: Since it's transcribed, maybe it's not invalid?

src/swap/changenow.js
```diff
@@ -314,6 +327,13 @@ export function makeChangeNowPlugin (
       const min = await get(
         'min-amount/' + quoteParams.from + '_' + quoteParams.to
       )
+      if (!min.minAmount) {
```
swansontec: This will trigger if `minAmount` is 0, which is still valid. I think it would be safer to have `min.minAmount != null`, so we only throw if the amount is really missing.

PR #50, opened 2019-10-18
Enable compound token exchange rates

kylanhurt:


src/rate/compound.js
```diff
@@ -0,0 +1,52 @@
...
+const codeTable = {}
+for (const number of Object.keys(currencies)) {
+  const entry = currencies[number]
+  codeTable[entry.Code] = true
+}
+
+function fixCurrency (currencyCode) {
+  currencyCode = currencyCode.toUpperCase()
+
+  return codeTable[currencyCode] ? 'iso:' + currencyCode : currencyCode
```
paullinator: This rate plug-in does not support any Fiat currencies. Therefore you do not need to look for any ISO Fiat currency codes. Please remove lines at nine through 13 and line 18 as they are not necessary. You can also remove the comment on line 36 as this is inaccurate

PR #62, opened 2020-02-28
Init sideshift plugin

sfncook:


swansontec (dismissed):
I had to rebase & adjust this a bit, so now it looks reasonable-ish. Testing will reveal any flaws.

PR #65, opened 2020-03-19
Integrate Switchain

dcusan:


src/swap/switchain.js
```diff
@@ -0,0 +1,371 @@
...
+      method
+    }
+
+    if (body) {
+      requestOpts = { ...requestOpts, body: JSON.stringify(body) }
+    }
+    let queryParams = ''
+    if (query) {
+      // only way i found to circumvent Object.entries type errors
+      const queryStringList: Array<[string, string]> = Object.entries(
```
swansontec: Our app needs to run in really old & bad Android WebView implementations that don't have these cool new `Object` methods. I suggest something more like:

```typescript
const queryStringList: Array<[string, string]> = []
for (const k of Object.keys(query)) {
  queryStringList.push([key, query[k]])
}
queryParams = `?${new URLSearchParams(queryStringList).toString()}`
```

Sorry.

src/swap/switchain.js
```diff
@@ -0,0 +1,371 @@
...
+          replyJson.status
+        } with JSON ${JSON.stringify(replyJson)}`
+      )
+    }
+
+    return replyJson
+  }
+
+  // closures to be populated between `fetchSwapQuote` and `approve`
+  let fromCurrencyCode = ''
```
swansontec: If you inline `makeSwapPluginQuote` into `fetchSwapQuote`, you won't need these closure variables. Might be cleaner, but I won't insist on a big refactor if you don't want to.

src/swap/switchain.js
```diff
@@ -0,0 +1,371 @@
...
+        fromNativeAmount = await fromWallet.denominationToNative(
+          div(add(toAmount, minerFee), quote, 8),
+          fromCurrencyCode
+        )
+        toNativeAmount = nativeAmount
+      }
+
+      const destinationAddress = toAddress
+      const expirationDate = new Date(expiryTs * 1000)
+      const isEstimate = false
```
swansontec: This is always a constant, so there is no need to pass it into your `makeSwapPluginQuote`.

src/swap/switchain.js
```diff
@@ -0,0 +1,371 @@
...
+
+      // create preliminary tx using our recieve address to calculate a networkFee
+      const spendInfo: EdgeSpendInfo = {
+        currencyCode: fromCurrencyCode,
+        spendTargets: [
+          { nativeAmount: fromNativeAmount, publicAddress: fromAddress }
```
swansontec: @peachbits Please test Ethereum with this. We might get "cannot send to your own address" errors on this line.

src/index.js
```diff
@@ -34,7 +35,8 @@ const edgeCorePlugins = {
   totle: makeTotlePlugin,
   foxExchange: makeFoxExchangePlugin,
   godex: makeGodexPlugin,
-  coinswitch: makeCoinSwitchPlugin
```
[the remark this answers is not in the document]
dcusan: I've sorted both lists.

yarn.lock
```diff
@@ -2,750 +2,770 @@
 # yarn lockfile v1
 
 
-"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.5.5":
```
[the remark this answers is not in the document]
dcusan: Reverted my yarn.lock changes.

src/swap/switchain.js
```diff
@@ -0,0 +1,371 @@
...
+  type EdgeSwapPlugin,
+  type EdgeSwapPluginQuote,
+  type EdgeSwapRequest,
+  type EdgeTransaction,
+  SwapAboveLimitError,
+  SwapBelowLimitError,
+  SwapCurrencyError
+} from 'edge-core-js/types'
+
+// import { makeSwapPluginQuote } from '../swap-helpers.js'
```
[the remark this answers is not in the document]
dcusan: Removed the comment.

src/swap/switchain.js
```diff
@@ -0,0 +1,371 @@
...
+  toAddressTag?: string,
+  refundAddress: string,
+  refundAddressTag?: string,
+  signature: string,
+  fromAmount?: string,
+  toAmount?: string
+}
+
+const dontUseLegacy = {
+  DGB: true
```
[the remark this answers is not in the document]
dcusan: I've added LTC to this object to derive addresses starting with `m...`

src/swap/switchain.js
dcusan: Thank you for the workaround!

src/swap/switchain.js
dcusan: Only the `orderCreationBody` closure remains as I need to bridge `fetchSwapQuote` and `makeSwapPluginQuote`.

src/swap/switchain.js
dcusan: Not entirely sure what you meant here :)

src/swap/switchain.js
dcusan: This is the only remaining issue. As this is only a temporary `EdgeSpendInfo`, is there a way to derive a new address to be used here?

Otherwise I can think of hardcoding an address for ETH and ERC20. Any other ideas>

dcusan:
Thanks for your comments, I've force pushed latest changes as per your request.

Only the issue with the same sending and receiving address remains. Please advise on how to proceed.

Thank you.

src/swap/switchain.js
swansontec: One of the parameters to `makeSwapPluginQuote` is `isEstimate`. However, `isEstimate` is always false. Therefore, you could just delete that parameter, and put the `false` directly inside `makeSwapPluginQuote`.

src/swap/switchain.js
swansontec: Using a hard-coded addresses is the only solution we've been able to come up with when we encounter this problem ourselves. Since Ethereum is balance-based, each wallet only has a single address, so there is no way to derive another one like we can in Bitcoin.

src/index.js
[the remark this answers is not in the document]
swansontec: Thanks.

src/swap/switchain.js
```diff
@@ -0,0 +1,365 @@
...
+      toNativeAmount,
+      networkFee: {
+        currencyCode: fromWallet.currencyInfo.currencyCode,
+        nativeAmount: tx.networkFee
+      },
+      destinationAddress,
+      pluginName: swapInfo.pluginName,
+      expirationDate,
+      quoteId,
+      isEstimate,
```
swansontec: This line could be `isEstimate: false,` removing the parameter. Also, you don't need the `pluginId` parameter either, now that I'm looking at it.

src/swap/switchain.js
```diff
@@ -0,0 +1,365 @@
...
+      async approve(): Promise<EdgeTransaction> {
+        const json: SwitchainOrderCreationResponse = await swHttpCall(
+          '/order',
+          'POST',
+          orderCreationBody
```
swansontec: Actually, this looks like a potential race condition, since we don't know that the `orderCreationBody` set up in the quote is necessarily the same one we want to approve (maybe the GUI gets a bug one day where it calls `quote` a bunch of times, leaving the wrong value in this variable and causing users to lose money). Could you pass this into `makeSwapPluginQuote` as a parameter, so we guarantee a 1:1 connection between quoting and approval?

dcusan: Moved `orderCreationBody` to `makeSwapPluginQuote` params as per your request.

swansontec:
All issues appear resolved in this latest version, so now it's down to testing.

swansontec:
I merged a rebased version of this, so it's all integrated. Thanks for your patience with this long review process.

PR #71, opened 2020-04-10
Matthew/swap promos

peachbits:


src/swap/changelly.js
```diff
@@ -182,8 +183,10 @@ export function makeChangellyPlugin(
 
     async getFixedQuote(
       request: EdgeSwapRequest,
-      userSettings: Object | void
+      userSettings: Object | void,
+      opts: { promoCode?: string }
     ): Promise<EdgeSwapQuote> {
+      –
```
swansontec, line 189: What is this line? Should be`const { promoCode } = opts`?

src/swap/changelly.js
```diff
@@ -142,7 +142,8 @@ export function makeChangellyPlugin(
     const headers = {
       'Content-Type': 'application/json',
       'api-key': apiKey,
-      sign
+      sign,
+      'X-Promo-Code': promoCode
```
swansontec, line 146: Do `if (promoCode != null) headers['X-Promo-Code'] = promoCode`

src/swap/changelly.js
```diff
@@ -143,8 +143,9 @@ export function makeChangellyPlugin(
       'Content-Type': 'application/json',
       'api-key': apiKey,
       sign,
-      'X-Promo-Code': promoCode
+      'X-Promo-Code': ''
```
swansontec, line 146: Delete this line. No need to send the header at all unless we have the code. That was the point of the `if`

PR #73, opened 2020-04-16
Matthew/swap fixes

peachbits:


src/swap/changenow.js
```diff
@@ -71,22 +71,22 @@ export function makeChangeNowPlugin(
   opts: EdgeCorePluginOptions
 ): EdgeSwapPlugin {
   const { initOptions, io, log } = opts
-  const { fetchCors = io.fetch } = io
+  const { fetch = io.fetch } = io
```
swansontec, line 74: We don't need the `= io.fetch` part.

PR #82, opened 2020-05-22
William/save swap data

swansontec:
This depends on https://github.com/EdgeApp/edge-core-js/pull/290, plus a bunch of GUI changes to stop writing out the old notes fields & show this new stuff.


swansontec:
Now that the core is published, this includes the proper version as a dependency.

swansontec:
Major version bump.

PR #87, opened 2020-07-08
Add FIO rate via BitMax

peachbits:


src/rate/bitmax.js
```diff
@@ -0,0 +1,46 @@
...
+            rate: 0.001
+          }
+        ]
+      }
+      asBitMaxTickerResponse(json)
+      return [
+        {
+          fromCurrency: 'FIO',
+          toCurrency: 'USDT',
+          rate: Number(json.data.close)
```
paullinator, line 41: Really you want the avg between the bid and ask. The `close` will give the previous day closing price. 

In summary, use the avg of `data.bid[0]` and `data.ask[0]`

peachbits: "close" is the last traded price and doesn't seem related to a time period https://bitmax-exchange.github.io/bitmax-pro-api/#ticker

PR #90, opened 2020-07-24
Enable Coingecko

kylanhurt:


.prettierrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
swansontec, line 1: We have a .prettierrc.json in edge-core-js that we should copy over here instead. We definintely don't want the `"printWidth": 120` line.

src/rate/coingecko.js
```diff
@@ -0,0 +1,34 @@
+// @flow
+
+import { type EdgeCorePluginOptions, type EdgeRatePlugin } from 'edge-core-js/types'
+
+export function makeCoinGeckoPlugin(opts: EdgeCorePluginOptions): EdgeRatePlugin {
+  const { io } = opts
+
+  return {
+    rateInfo: {
+      displayName: 'Coincap',
```
swansontec, line 10: CoinGecko?

src/rate/coingecko.js
```diff
@@ -0,0 +1,34 @@
...
+
+    async fetchRates(pairsHint) {
+      const reply = await io.fetch(
+        'https://api.coingecko.com/api/v3/coins/telos?localization=false&tickers=false&market_data=true&community_data=false&developer_data=false&sparkline=false'
+      )
+      const json = await reply.json()
+
+      // Grab all the pairs (TLOS to _______)
+      const pairs = []
+      if (json && json.market_data && json.market_data.current_price && json.market_data.current_price.usd) {
```
swansontec, line 21: Are you familiar with cleaners? We generally prefer those for JSON validation going forward.

```
import { asObject, asNumber } from 'cleaners'

type asGeckoUsdReply = asObject({
  market_data: asObject({
    current_price: asObject({
      usd: asNumber
    })
  })
})
```

This is like a TypeScript type definition, but it happens at runtime. Now we can use this function to validate the JSON:

```
try {
  const rate = asGeckoUsdReply(json).market_data.current_price.usd
  pairs.push({ ..., rate })
} catch () {}
```

src/index.js
```diff
@@ -29,6 +30,7 @@ const edgeCorePlugins = {
   coinmonitor: makeCoinmonitorPlugin,
   compound: makeCompoundPlugin,
   constantRate: makeConstantRatePlugin,
+  coingecko: makeCoinGeckoPlugin,
```
swansontec, line 33: Please alphabetize this.

src/index.js
```diff
@@ -47,10 +49,7 @@ const edgeCorePlugins = {
   transfer: makeTransferPlugin
 }
 
-if (
-  typeof window !== 'undefined' &&
-  typeof window.addEdgeCorePlugins === 'function'
-) {
+if (typeof window !== 'undefined' && typeof window.addEdgeCorePlugins === 'function') {
```
swansontec, line 52: This line should be the way it was, before the weird prettier file messed it up.

PR #99, opened 2020-10-08
Update Changelly to getFixRateForAmount

peachbits:
New method doesn't need min/max checking since the Changelly response indicates if the trade is possible or not rather than returning acceptable amounts for a range pair. Error code is generic and the error message string would need to be parsed to determine which limit wasn't met

### DIFF
--- a/src/swap/changelly.js
+++ b/src/swap/changelly.js
@@ -1,6 +1,6 @@
 // @flow
 
-import { gt, lt, mul } from 'biggystring'
+import { lt, mul } from 'biggystring'
 import {
   type EdgeCorePluginOptions,
   type EdgeCurrencyWallet,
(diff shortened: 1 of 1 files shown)


src/swap/changelly.js
```diff
@@ -213,37 +225,17 @@ export function makeChangellyPlugin(
       if (CURRENCY_CODE_TRANSCRIPTION[request.toCurrencyCode]) {
         safeToCurrencyCode = CURRENCY_CODE_TRANSCRIPTION[request.toCurrencyCode]
       }
-      const fixedRateQuote = await call({
+      const fixedRateQuote: FixedRateQuote = await call({
```
paullinator, line 228: If you are going to add a type, might as well use cleaners.

paullinator (changes requested):

paullinator (approved):

PR #101, opened 2020-11-13
Implement rate hints

peachbits:


src/rate/coinbase.js
```diff
@@ -30,24 +24,24 @@ export function makeCoinbasePlugin(
...
-      for (const key of keys) {
-        const rate = Number(json.data.rates[key])
-        const toCurrency = fixCurrency(key)
-        pairs.push({
-          fromCurrency: 'iso:USD',
-          toCurrency,
-          rate
-        })
+      for (const pair of pairsHint) {
+        const fiatCode = pair.toCurrency.split(':')
```
paullinator, line 29: This will cost a lot of fetches. Better to do the single fetch that returns a lot of pairs but only filter the pairs that are in the hints

src/rate/coincap.js
```diff
@@ -15,30 +31,41 @@ export function makeCoincapPlugin(opts: EdgeCorePluginOptions): EdgeRatePlugin {
...
+      if (Object.keys(currencyMap).length === 0) {
+        const assets = await io.fetch(`https://api.coincap.io/v2/assets/`)
+        const assetsJson = await assets.json()
+        const assetIds = asCoincapAssets(assetsJson.data)
+        assetIds.forEach(code => (currencyMap[code.symbol] = code.id))
+        log(`rate data coincap ${JSON.stringify(currencyMap)}`)
+      }
+      for (const pair of pairsHint) {
+        // Coincap only provides prices in USD and must be queried by unique identifier rather that currency code
+        if (!currencyMap[pair.fromCurrency] || pair.toCurrency !== 'iso:USD')
```
paullinator, line 44: It is still worth returning The pair even if it's not USD. The core can find A non USD rate by routing through A fiat exchange

paullinator (changes requested):

paullinator (approved):

PR #113, opened 2021-01-26
Move aTokens into constantRate plugin

peachbits:


paullinator:
Im confused. I thought the purpose was to remove exchange rates entirely as the constant rate does not technically match the underlying assets.

paullinator (approved):

PR #118, opened 2021-03-08
Fix min/max error for SideShift.ai

szgergely:
This PR fixes the issue where SideShift.ai plugin throws `No enabled exchanges support` error when the exchange amount is below the minimum/above maximum, instead of displaying the minimum/maximum.

[Support ticket](https://support.edge.app/public/tickets/bd338d1181db2cd63e5206f59a743998d0376d7757e8bc9aba651e43946d283e)


src/swap/sideshift.js
```diff
@@ -185,7 +185,7 @@ const createFetchSwapQuote = (api: SideshiftApi, affiliateId: string) =>
     )
 
     if (fixedQuote.error) {
-      throw checkQuoteError(rate, request, fixedQuote.error.message)
+      await checkQuoteError(rate, request, fixedQuote.error.message)
```
peachbits, line 188: This fixes the reported issue but creates a handful of flow errors since flow isn't seeing that checkQuoteError always throws. To get around that limitation, and avoid ugly flow hacks, let's add the generic throw from line 92 right after this line

src/swap/sideshift.js
```diff
@@ -200,7 +200,7 @@ const createFetchSwapQuote = (api: SideshiftApi, affiliateId: string) =>
     )
 
     if (order.error) {
-      throw checkQuoteError(rate, request, order.error.message)
+      await checkQuoteError(rate, request, order.error.message)
```
peachbits, line 203: Same as comment above

peachbits (changes requested):

peachbits (approved):

PR #141, opened 2021-07-30
Fix currency converter response cleaners

peachbits:


thehobbit85 (approved):

thehobbit85 (changes requested):
You need to keep the 'error' is it exists

thehobbit85 (approved):

PR #152, opened 2021-10-05
Sideshift: include destinationTag in XRP tx

szgergely:
At the moment XRP transactions sent to SideShift.ai do not include the `destinationTag` returned by the SideShift API, therefore they are rejected by the XRP network.

This PR appends the `destinationTag` to the XRP transactions, so they will be processed successfully.

### DIFF
--- a/src/swap/sideshift.js
+++ b/src/swap/sideshift.js
@@ -1,6 +1,13 @@
 // @flow
 
-import { asBoolean, asEither, asObject, asOptional, asString } from 'cleaners'
+import {
+  asBoolean,
+  asEither,
+  asNumber,
+  asObject,
+  asOptional,
+  asString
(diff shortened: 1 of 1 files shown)


src/swap/sideshift.js
```diff
@@ -235,13 +242,17 @@ const createFetchSwapQuote = (api: SideshiftApi, affiliateId: string) =>
 
     const isEstimate = false
 
+    const destinationTag = order.depositAddress.destinationTag
```
thehobbit85, line 245: FlowJS calls it [sketchy-null](https://github.com/facebook/flow/blob/main/website/en/docs/linting/rule-reference.md#sketchy-null-)

thehobbit85 (approved):

PR #154, opened 2021-11-24
Use the correct currencyCode in the metadata

thehobbit85:
---
  - https://app.asana.com/0/0/1201411731543809


thehobbit85:
Approved OTS by @paullinator

PR #170, opened 2022-02-02
letsexchange - add plugin

xcodex2010:
letsexchange - add plugin

### DIFF
--- a/src/swap/letsexchange.js
+++ b/src/swap/letsexchange.js
@@ -29,7 +29,7 @@ const swapInfo: EdgeSwapInfo = {
   supportEmail: 'support@letsexchange.io'
 }
 
-const orderUri = 'https://api.letsexchange.io/?exchangeId='
+const orderUri = 'https://letsexchange.io/?exchangeId='
 const uri = 'https://api.letsexchange.io/api/v1/'
 
 const expirationMs = 1000 * 60 * 20
--- a/src/swap/letsexchange.js
+++ b/src/swap/letsexchange.js
@@ -1,5 +1,13 @@
 // @flow
 
+import {
+  asBoolean,
+  asEither,
+  asNumber,
+  asObject,
+  asOptional,
+  asString
+} from 'cleaners'
 import { lt } from 'biggystring'
 import {
   type EdgeCorePluginOptions,
@@ -33,28 +41,27 @@ const uri = 'https://api.letsexchange.io/api/v1/'
 
 const expirationMs = 1000 * 60 * 20
 
-type QuoteInfo = {
-  transaction_id: string,
-  status: string,
-  coin_from: string,
-  coin_to: string,
-  network_from: string,
-  network_to: string,
-  deposit_amount: string,
-  withdrawal_amount: string,
-  deposit: string,
-  deposit_extra_id: string,
-  withdrawal: string,
-  withdrawal_extra_id: string,
-  rate: string,
-  fee: string,
-  return: string,
-  return_extra_id: string,
-  final_amount: string,
-  hash_in: string,
-  hash_out: string,
-  isEstimate: boolean
-}
+const asQuoteInfo = asObject({
+  transaction_id: asString,
+  status: asString,
+  coin_from: asString,
+  coin_to: asString,
+  network_from: asString,
+  network_to: asString,
+  deposit_amount: asString,
+  withdrawal_amount: asString,
+  deposit: asString,
+  deposit_extra_id: asString,
+  withdrawal: asString,
+  withdrawal_extra_id: asString,
+  rate: asString,
+  fee: asString,
+  return: asString,
+  final_amount: asString,
+  hash_in: asString,
+  hash_out: asString,
+  isEstimate: asBoolean,
+})
 
 const dontUseLegacy = {
   DGB: true
@@ -207,8 +214,6 @@ export function makeLetsExchangePlugin(
           network_to: toMainnetCode,
           withdrawal: toAddress,
           return: fromAddress,
-          // return_extra_id: 'empty',
-          // withdrawal_extra_id: 'empty',
           return_extra_id: null,
           withdrawal_extra_id: null,
           affiliate_id: initOptions.apiKey,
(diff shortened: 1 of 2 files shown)


src/swap/letsexchange.js
```diff
@@ -0,0 +1,294 @@
...
+      break;
+    case 'BSC':
+      networkCode = 'BEP20';
+      break;
+    default:
+      break;
+  }
+
+  return networkCode;
+}
```
thehobbit85, line 104: Don't use a 'switch case' function, use the 'getCodesWithMainnetTranscription' instead.
Here is the link to the function: https://github.com/EdgeApp/edge-exchange-plugins/blob/80fac2c1ca153234f613241e4a5e7ef788c2a731/src/swap-helpers.js#L188
And this is an example usage: https://github.com/EdgeApp/edge-exchange-plugins/blob/80fac2c1ca153234f613241e4a5e7ef788c2a731/src/swap/changenow.js#L105

thehobbit85 (changes requested):

src/swap/letsexchange.js
```diff
@@ -0,0 +1,294 @@
...
+    const response = await fetchCors(url, { method: 'POST', body, headers })
+    if (!response.ok) {
+      if (response.status === 422) {
+        throw new SwapCurrencyError(
+          swapInfo,
+          request.fromCurrencyCode,
+          request.toCurrencyCode
+        )
+      }
+      throw new Error(`letsexchange returned error code ${response.status}`)
```
paullinator, line 128: Are there any cases where the user might get an "over limit" error or "region not supported" error? If so, need to catch these and throw the appropriate error.

src/swap/letsexchange.js
```diff
@@ -0,0 +1,294 @@
...
+          withdrawal_extra_id: null,
+          affiliate_id: initOptions.apiKey,
+          promocode: promoCode != null ? promoCode : '',
+          type: 'edge',
+          float: true,
+          isEstimate: false
+        }
+      })
+      log('sendReply' + sendReply)
+      const quoteInfo: QuoteInfo = sendReply
```
paullinator, line 246: This should use a `cleaner`. See https://github.com/EdgeApp/edge-exchange-plugins/blob/master/src/swap/sideshift.js#L377

paullinator (changes requested):

src/swap/letsexchange.js
xcodex2010: Is there an example?

src/swap/letsexchange.js
xcodex2010: Fixed it.

src/swap/letsexchange.js
xcodex2010: No. There are not.

src/swap/letsexchange.js
paullinator: The link has an example

xcodex2010: Fixed it.

src/swap/letsexchange.js
```diff
@@ -0,0 +1,296 @@
...
+          promocode: promoCode != null ? promoCode : '',
+          type: 'edge',
+          float: true,
+          isEstimate: false
+        }
+      })
+
+      log('sendReply' + sendReply)
+      const quoteInfo = asQuoteInfo({
+        transaction_id: sendReply.transaction_id,
```
paullinator, line 229: Not the right way to use a cleaner. You should take the raw data coming from an api call and pass it through a cleaner to validate the response.

ie 

```
const sendReply = await call(uri + 'transaction', request
const quoteInfo = asQuoteInfo(sendReply)
```

paullinator (changes requested):

src/swap/letsexchange.js
```diff
@@ -0,0 +1,296 @@
...
+} from '../swap-helpers.js'
+
+const pluginId = 'letsexchange'
+const swapInfo: EdgeSwapInfo = {
+  pluginId,
+  displayName: 'LetsExchange',
+  supportEmail: 'support@letsexchange.io'
+}
+
+const orderUri = 'https://api.letsexchange.io/?exchangeId='
```
paullinator, line 39: This URL does not bring the user to a valid order status page.

paullinator (changes requested):

src/swap/letsexchange.js
xcodex2010: this url show exchange info page, like this
image

src/swap/letsexchange.js
xcodex2010: Thank you! Got it. 
Done.

src/swap/letsexchange.js
paullinator: That was not my experience in testing. The links go to an API doc page.

paullinator:
App crashes in testing. 

yarn precommit found this issue causing the crash.
```Monitor logs will go to /private/tmp/flow/zSUserszSpaulzSgitzSedge-exchange-plugins.monitor_log
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/swap/letsexchange.js:29:3

Cannot import getCodesWithMainnetTranscription because there is no getCodesWithMainnetTranscription export in
../swap-helpers.js.

     26│   type InvalidCurrencyCodes,
     27│   checkInvalidCodes,
     28│   makeSwapPluginQuote, getCodes,
     29│   getCodesWithMainnetTranscription
     30│ } from '../swap-helpers.js'
     31│
     32│ const pluginId = 'letsexchange'



Found 1 error
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
paul@Macallan edge-exchange-plugins % 
```
Please thoroughly test the plugin before submitting the PR.

paullinator (changes requested):
Fix crash and thoroughly test

paullinator:
please do not merge master in to you feature branch. makes it difficult to find changes. Please rebase your commits on top of master instead. As well, squash all commits into one as this is a small PR.

paullinator:
in testing i could not swap BTC to ETH. Got this error
Screen Shot 2022-03-25 at 18 57 38

xcodex2010:
Hi! Tell me please. How could i testing it without mobile app ?
Is it possible ?

xcodex2010:
perhaps this is possible with **test/demo.js** ?

paullinator:
No. You need the mobile app. See readme

xcodex2010:
Hi!
Is there the step by step instruction for run and test exchange app ?
I did all that describe at readme. Run app via xcode emulator.
on edge-react-gui ran yarn ios. that ran app but without any exchanges at Settings > Exchange Settings. And could i get fake money on my test wallet at my app?

xcodex2010:
Hi!
where can I get the log data that I describe in edge-exchange-plugins using the log() method;
How to debug my (edge-exchange-plugins) plugin ?

xcodex2010:
Hi!. All done! Has been successfully tested! 
image

paullinator:
Please do not create any merge commits into your feature branch. Please fix the PR and squash all commits into one with no merge commit. Instead of creating a merge from master to your branch, rebase your branch on top of master

paullinator (changes requested):
see comment

xcodex2010:
Sorry. I created new PR. Because this PR is like a trash already.

https://github.com/EdgeApp/edge-exchange-plugins/pull/183

xcodex2010:
image

Does it matter? Or can we send it to you in another way ?

PR #173, opened 2022-02-16
Update Exolix to use mainnet:token params

peachbits:
I reached out to Exolix regarding ZEC z-address support and will update the PR if they respond before the next publish


paullinator (approved):

peachbits:
Confirmed that Exolix only supports t-addresses so we will block ZEC purchases for now.

PR #180, opened 2022-03-24
ChangeHero plugin

pavelks-dev:
Changehero plugin. Copy of changelly plugin with a few essential changes.

### DIFF
--- a/src/swap/changehero.js
+++ b/src/swap/changehero.js
@@ -1,0 +1,281 @@
+// @flow
+
+import { asNumber, asObject, asOptional, asString } from 'cleaners'
+import {
+  type EdgeCorePluginOptions,
+  type EdgeCurrencyWallet,
+  type EdgeSpendInfo,
+  type EdgeSwapInfo,
+  type EdgeSwapPlugin,
+  type EdgeSwapQuote,
+  type EdgeSwapRequest,
+  type EdgeTransaction,
+  SwapCurrencyError
+} from 'edge-core-js/types'
+
+import { makeSwapPluginQuote, safeCurrencyCodes } from '../swap-helpers.js'
+
+const CURRENCY_CODE_TRANSCRIPTION = {
+  ethereum: {
+    USDT: 'USDT20'
+  },
+  avalanche: {
+    AVAX: 'AVAXC'
+  },
+  binancesmartchain: {
+    BNB: 'BNBBSC'
+  },
+  polygon: {
+    MATIC: 'POLYGON'
+  }
+}
+
+const pluginId = 'changehero'
+const swapInfo: EdgeSwapInfo = {
+  pluginId,
+  displayName: 'ChangeHero',
+  supportEmail: 'support@changehero.io'
+}
+
+const orderUri = 'https://changehero.io/transaction/'
+const uri = 'https://api.changehero.io/v2'
+const expirationFixedMs = 1000 * 60 * 5
+
+type FixedQuoteInfo = {
+  id: string,
+  amountExpectedFrom: string,
+  amountExpectedTo: number,
+  amountTo: number,
+  createdAt: string,
+  currencyFrom: string,
+  currencyTo: string,
+  payinAddress: string,
+  payinExtraId: string | null,
+  payoutAddress: string,
+  payoutExtraId: string | null,
+  refundAddress: string,
+  refundExtraId: string | null,
+  status: string
+}
+
+const asQuoteInfo = asObject({
+  result: asObject({
+    id: asString,
+    status: asString,
+    amountExpectedFrom: asString,
+    amountExpectedTo: asNumber,
+    payinAddress: asString,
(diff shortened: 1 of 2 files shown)


src/swap/changehero.js
```diff
@@ -0,0 +1,412 @@
...
+            }
+
+      const sendReply = await call({
+        jsonrpc: '2.0',
+        id: 2,
+        method: 'createFixTransaction',
+        params
+      })
+      checkReply(sendReply, request)
+      const quoteInfo: FixedQuoteInfo = sendReply.result
```
paullinator, line 207: results from servers should use `cleaners` to check the payload at runtime. See

https://github.com/EdgeApp/edge-exchange-plugins/blob/master/src/swap/letsexchange.js#L38

src/swap/changehero.js
```diff
@@ -0,0 +1,412 @@
...
+        method: 'getFixRate',
+        params: {
+          from: safeFromCurrencyCode,
+          to: safeToCurrencyCode
+        }
+      })
+
+      const [{ id: responseId }] = fixedRateQuote.result
+      const params =
+        request.quoteFor === 'from'
```
paullinator, line 178: API does not seem to be able to specify the network chain separately from the token code. This is mandatory for integration. See the LetsExchange plugin here

https://github.com/EdgeApp/edge-exchange-plugins/blob/master/src/swap/letsexchange.js#L160

paullinator (changes requested):
Not a complete review, but API does not sufficiently distinguish chain vs token code. This must be resolved before continuing.

src/swap/changehero.js
```diff
@@ -0,0 +1,411 @@
...
+            uniqueIdentifier: quoteInfo.payinExtraId || undefined
+          }
+        ],
+        networkFeeOption:
+          request.fromCurrencyCode.toUpperCase() === 'BTC'
+            ? 'high'
+            : 'standard',
+        swapData: {
+          orderId: quoteInfo.id,
+          isEstimate: false,
```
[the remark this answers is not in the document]
p-cse: added the main page https://changehero.io

paullinator:
Please rebuild this PR as one commit. Do not merge master into this branch. If needed, rebase this branch on top of master. That’s our current workflow

src/swap/changehero.js
```diff
@@ -0,0 +1,411 @@
...
+    swapInfo,
+    async fetchSwapQuote(
+      request: EdgeSwapRequest,
+      userSettings: Object | void
+    ): Promise<EdgeSwapQuote> {
+      const fixedPromise = this.getFixedQuote(request, userSettings)
+      // Estimated swaps disabled
+      const fixedResult = await fixedPromise
+      return fixedResult
+    },
```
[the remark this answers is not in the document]
p-cse: fixed

src/swap/changehero.js
p-cse: Do you mean that we need to add parameters network_from and network_to to our api support?

magmel48: we specify network in tickers usually, that's our current idea. Also I see that another exchangers use INVALID_CURRENCY_CODES, for instance here https://github.com/EdgeApp/edge-exchange-plugins/blob/master/src/swap/changelly.js#L28 and no network in changelly plugin.

paullinator: Yes Changelly doesn’t have network support but we are not allowing new plugins to not have an explicit network field. As well we’ll be deprecating any plug-ins that don’t have that feature. It causes too many support issues and apparent loss of funds.

magmel48: But if you need to exchange for instance USDC on MATIC mainnet you have to use not USDC ticker, but USDCPOLYGON. That's our idea. Is there any way to specify some logic of such conversion from USDC to USDCPOLYGON in exchange plugin, but not on ChangeHero API side?

paullinator: While that’s doable, it’s a hack and doesn’t scale to support a lot of tokens. As well we would have to update the app every time you add new tokens. 

There can also be errors by which a new chain also supports USDC but you haven’t added a ticker conversion for it and the user then receives ETH USDC.

magmel48: That's tricky for us to implement at the current times. Is there any way to disable all chains except ETH to proceed with the release while we are finding a solution? Is there any way to deliver such functionality in the future Edge wallet releases?

paullinator: Sure. if you compare the `fromWallet.currencyInfo.currencyCode` to the `fromCurrencyCode` (and do the same for the `to` values, they should be the same and can only be different if `fromWallet.currencyInfo.currencyCode` is `ETH`

p-cse: Look please. I added checkEthOnly function to check ETH tokens

paullinator:
Replaced by #198

PR #183, opened 2022-04-06
LetsExchange - exchange plugin

xcodex2010:


src/swap/letsexchange.js
```diff
@@ -0,0 +1,268 @@
...
+
+// Network names that don't match parent network currency code
+const MAINNET_CODE_TRANSCRIPTION = {
+  ethereum: 'ERC20',
+  binancesmartchain: 'BEP20',
+  tron: 'TRC20',
+  bnb: 'BEP2'
+}
+
+async function getAddress(wallet: EdgeCurrencyWallet, currencyCode: string) {
```
peachbits, line 86: Missing type annotation

src/swap/letsexchange.js
```diff
@@ -0,0 +1,268 @@
...
+          network_from: fromMainnetCode,
+          network_to: toMainnetCode,
+          withdrawal: toAddress,
+          return: fromAddress,
+          return_extra_id: null,
+          withdrawal_extra_id: null,
+          affiliate_id: initOptions.apiKey,
+          promocode: promoCode != null ? promoCode : '',
+          type: 'edge',
+          float: true,
```
peachbits, line 213: You've specified a floating exchange rate here but in the spendInfo isEstimate is hardcoded to false. For a floating rate exchange,  isEstimate should be true.

src/swap/letsexchange.js
```diff
@@ -0,0 +1,268 @@
...
+          coin_to: request.toCurrencyCode,
+          network_from: fromMainnetCode,
+          network_to: toMainnetCode,
+          withdrawal: toAddress,
+          return: fromAddress,
+          return_extra_id: null,
+          withdrawal_extra_id: null,
+          affiliate_id: initOptions.apiKey,
+          promocode: promoCode != null ? promoCode : '',
+          type: 'edge',
```
peachbits, line 212: This likely doesn't matter but worth noting I didn't see type or isEstimate in the LetsExchange docs yet they are included in the payload here.

src/swap/letsexchange.js
```diff
@@ -0,0 +1,268 @@
...
+  hash_out: asOptionalBlank(asString)
+})
+
+const dontUseLegacy = {
+  DGB: true
+}
+
+const INVALID_CURRENCY_CODES: InvalidCurrencyCodes = {
+  from: {
+    ETH: ['MATIC'],
```
peachbits, line 64: It appears the LetsExchange API supports identifying the mainnet currency when creating orders so we can loosen up the limits here. Another plugin that can similarly use mainnet codes is Changenow you can likely use a similar list of invalid codes. 

A note about ZEC, Edge only creates shielded addresses so if LetsExchange cannot send to a shielded address than it should remain invalid here.

src/swap/letsexchange.js
```diff
@@ -0,0 +1,268 @@
...
+          }
+        ],
+        networkFeeOption:
+          request.fromCurrencyCode.toUpperCase() === 'BTC'
+            ? 'high'
+            : 'standard',
+        swapData: {
+          orderId: quoteInfo.transaction_id,
+          orderUri: orderUri + quoteInfo.transaction_id,
+          isEstimate: false,
```
peachbits, line 238: Another note on floating and fixed exchanges, if LetsExchange supports different limits for each (like lower minimum amounts for floating rate) then it's worth attempting both types. For example, if LetsExchange can't server a fixed rate swap try again specifying a floating rate.

peachbits (changes requested):

src/swap/letsexchange.js
xcodex2010: Done!

src/swap/letsexchange.js
xcodex2010: Done!

src/swap/letsexchange.js
xcodex2010: Done!

src/swap/letsexchange.js
xcodex2010: Done!

src/swap/letsexchange.js
xcodex2010: Done!

src/swap/letsexchange.js
```diff
@@ -0,0 +1,262 @@
...
+const INVALID_CURRENCY_CODES: InvalidCurrencyCodes = {
+  from: {},
+  to: {
+    zcash: ['ZEC']
+  }
+}
+
+// Network names that don't match parent network currency code
+const MAINNET_CODE_TRANSCRIPTION = {
+  ethereum: 'ERC20',
```
peachbits, line 71: I tested the plugin and I wasn't able to get a quote for BTC -> ETH and it failed on the /info call. It appears the issue is the network_to and network_from fields have multiple possible in the LetsExchange api for each network. For example, when trading for ETH the network_to is supposed to be 'ETH' but when trading an ERC20 token like USDT the network_to is supposed to be 'ERC20'. There will need to be logic to determine if the main chain currency code or it's token type should be used in the network fields.

src/swap/letsexchange.js
```diff
@@ -0,0 +1,262 @@
...
+          network_from: fromMainnetCode,
+          network_to: toMainnetCode,
+          withdrawal: toAddress,
+          return: fromAddress,
+          return_extra_id: null,
+          withdrawal_extra_id: null,
+          affiliate_id: initOptions.apiKey,
+          promocode: promoCode != null ? promoCode : '',
+          type: 'edge',
+          float: true,
```
peachbits, line 207: I previously requested that the plugin attempt both types of trade, fixed and floating, as this will maximize the opportunities for LetsExchange to be presented to a user. If you're only going to go with a single type I highly recommend doing a fixed rate since the app will always prioritize fixed rate trades over a floating rate.

peachbits (changes requested):

src/swap/letsexchange.js
xcodex2010: Fixed it. But I didn't find ETH in the app.
image width=200px

peachbits: There was a recent bug affecting search results I suspect you're running a version of edge-react-gui that still has it (issue was fixed ~10 days ago). 

The change you made doesn't address the issue I raised. I was describing how the LetsExchange API works and the logic necessary to accurately request quotes. For example:

1) User wants to trade ETH so the relevant network field needs to be ETH.
2) User wants to trade USDT (on ethereum) so the relevant network field needs to be ERC20

The logic needs to be, if the currency being traded is the main chain currency, then use that currency code as the network (you can use request.fromWallet.currencyInfo.currencyCode and request.toWallet.currencyInfo.currencyCode to do the comparison). Else, it's a token and you'll need to use the MAINNET_CODE_TRANSCRIPTION map to get the appropriate network name. 

Revert the change you make and apply the logic above and you should be good to go.

xcodex2010: Hi! I did a "git pull" for edge-react-gui. My wallet list got bigger but ETH is still missing.

xcodex2010: How can I select USDT (on Ethereum) or any another coin on Ethereum in the app?
I only see the name of the coin, token. I don't see the network name

xcodex2010: This is what i got when i chose USDT(on AVAX)
image

xcodex2010: This is what i got when try to chose USDT(on AVAX)

image

xcodex2010: I did new PR with all fixes.
https://github.com/EdgeApp/edge-exchange-plugins/pull/190

peachbits:
Closed in favor of https://github.com/EdgeApp/edge-exchange-plugins/pull/190

PR #192, opened 2022-05-23
Add api2 criptointercambio

CriptointercambioDev:


src/swap/criptointercambio.js
```diff
@@ -0,0 +1,488 @@
...
+  const out: EdgeSwapPlugin = {
+    swapInfo,
+    async fetchSwapQuote(
+      request: EdgeSwapRequest,
+      userSettings: Object | void,
+      opts: { promoCode?: string }
+    ): Promise<EdgeSwapQuote> {
+      checkInvalidCodes(INVALID_CURRENCY_CODES, request, swapInfo)
+
+      const fixedPromise = this.getFixedQuote(request, userSettings, opts)
```
paullinator, line 199: just return here. No need for the `const fixedPromise`

src/swap/criptointercambio.js
```diff
@@ -0,0 +1,488 @@
...
+          ? await request.fromWallet.nativeToDenomination(
+            request.nativeAmount,
+            request.fromCurrencyCode
+          )
+          : await request.toWallet.nativeToDenomination(
+            request.nativeAmount,
+            request.toCurrencyCode
+          )
+
+      const { safeFromCurrencyCode, safeToCurrencyCode } = safeCurrencyCodes(
```
paullinator, line 226: This is a bit concerning. Your native API should support specifying the chaincode separately from the token code. Without that, there could easily be confusion on what token the user wants to send or purchase. 

The CURRENCY_CODE_TRANSCRIPTION is a quick hack for prior partners but all new partners should have an explicit chaincode and tokencode. See the Changenow plugin

https://github.com/EdgeApp/edge-exchange-plugins/blob/master/src/swap/changenow.js#L110

paullinator (changes requested):

src/swap/criptointercambio.js
```diff
@@ -0,0 +1,488 @@
...
+      )
+
+      const fixedRateQuoteResponse: FixedRateQuote = await call({
+        jsonrpc: '2.0',
+        id: 'one',
+        method: 'getFixRateForAmount',
+        params: {
+          from: safeFromCurrencyCode,
+          to: safeToCurrencyCode,
+          amountFrom: quoteAmount
```
paullinator, line 238: `quoteAmount` is not always the FROM amount. Need to check the `request.quoteFor` value and check if it's `from` or `to`. If you can't do quotes coming from `to` amounts, then you need to throw.

src/swap/criptointercambio.js
```diff
@@ -0,0 +1,488 @@
...
+        }
+      })
+      const fixedRateQuote = asFixedRateQuote(fixedRateQuoteResponse)
```
paullinator, line 241: Need to check limits somewhere. The code doesn't throw the `SwapAboveLimitError` or `SwapBelowLimitError` at all. See the changenow plugin

PR #195, opened 2022-06-16
Changelly: Re-enable estimated rate swaps

peachbits:
---
  - https://app.asana.com/0/0/1202288034748122

### DIFF
--- a/src/swap/changelly.js
+++ b/src/swap/changelly.js
@@ -199,9 +199,14 @@ export function makeChangellyPlugin(
       checkInvalidCodes(INVALID_CURRENCY_CODES, request, swapInfo)
 
       const fixedPromise = this.getFixedQuote(request, userSettings, opts)
-      // FIXME: Estimated swaps are temporarily disabled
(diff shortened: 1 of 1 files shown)


src/swap/changelly.js
```diff
@@ -199,9 +199,14 @@ export function makeChangellyPlugin(
...
-      const fixedResult = await fixedPromise
-      return fixedResult
+      const estimatePromise = this.getEstimate(request, userSettings, opts)
+      try {
+        const fixedResult = await fixedPromise
+        return fixedResult
+      } catch (e) {
+        const estimateResult = await estimatePromise
+        return estimateResult
+      }
```
thehobbit85, line 209: Not that it matters or anything, but this is kinda gross, this could have just be

```js 
try {
   return await this.getFixedQuote(request, userSettings, opts)
} catch (e) {
   return await this.getEstimate(request, userSettings, opts)
}
```
You don't really have to change it since it doesn't really matter, but I had to mention it ;)

thehobbit85 (approved):

PR #196, opened 2022-06-21
Remove Changelly

thehobbit85:


src/swap/changenow.js
```diff
@@ -118,17 +118,15 @@ export function makeChangeNowPlugin(
...
       ): Promise<ChangeNowResponse> {
         const type = isSelling ? 'direct' : 'reverse'
-
-        // Get rateId and Date
-        const exchangeAmountResponse = await fetch(
+        const fetchUrl =
           uri +
-            `exchange/estimated-amount?flow=${flow}&useRateId=${String(
-              flow === 'fixed-rate'
-            )}&${
```
paullinator, line 127: Why was this included?

paullinator (changes requested):

src/swap/changenow.js
thehobbit85: I have no idea ¯\_(ツ)_/¯
I didn't noticed it was part of the commit and I don't even remember ever making this change. 

I'll remove it.

paullinator (approved):

PR #199, opened 2022-07-10
Add Cryptointercambio

paullinator:


paullinator:
@CriptointercambioDev Please look at my fixup commit. Also, the original PR is missing some proper error handling. You need to properly detect when the requested amount is above or below the limit and return the min/max amount accordingly using the SwapAboveLimitError or SwapBelowLimitError. As well, if there are any georestriction, return the SwapPermissionError. See https://github.com/EdgeApp/edge-exchange-plugins/blob/1396f800bc11db2fa918a4af0af07826bfa6de5a/src/swap/sideshift.js for examples

PR #204, opened 2022-07-30
Do not allow ChangeHero to swap tokens that are EVM chains

paullinator:
---
  - https://app.asana.com/0/0/1202687715819288
  - https://app.asana.com/0/0/1202692200023371

### DIFF
--- a/src/swap/changehero.js
+++ b/src/swap/changehero.js
@@ -23,7 +23,12 @@ import {
   SwapCurrencyError
 } from 'edge-core-js/types'
 
-import { makeSwapPluginQuote, safeCurrencyCodes } from '../swap-helpers.js'
+import {
+  type InvalidCurrencyCodes,
+  checkInvalidCodes,
+  makeSwapPluginQuote,
+  safeCurrencyCodes
+} from '../swap-helpers.js'
 
 const CURRENCY_CODE_TRANSCRIPTION = {
   ethereum: {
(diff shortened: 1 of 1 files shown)


thehobbit85 (changes requested):
We actually already have a mechanism for this in place: 
https://github.com/EdgeApp/edge-exchange-plugins/blob/ec8a6435ec68bd51473b84c9f83ac1499848e76a/src/swap/switchain.js#L73

https://github.com/EdgeApp/edge-exchange-plugins/blob/ec8a6435ec68bd51473b84c9f83ac1499848e76a/src/swap/switchain.js#L200

Can you change it to use that convention?

src/swap/changehero.js
```diff
@@ -40,6 +45,25 @@ const CURRENCY_CODE_TRANSCRIPTION = {
...
+    fantom: 'allCodes',
+    polygon: 'allCodes'
+  },
+  to: {
+    ethereum: ['MATIC', 'AVAX', 'BNB', 'FTM', 'CELO'],
+    avalanche: 'allTokens',
+    binancesmartchain: 'allCodes',
+    celo: 'allTokens',
+    fantom: 'allCodes',
+    polygon: 'allCodes'
```
thehobbit85, line 63: Change all to 'allCodes'

thehobbit85 (changes requested):

thehobbit85 (approved):

PR #205, opened 2022-08-15
Api updated to v2

ProcyonLotor123:
---
  - https://app.asana.com/0/0/1204464594667832

### DIFF
--- a/src/swap/exolix.ts
+++ b/src/swap/exolix.ts
@@ -1,5 +1,12 @@
 import { lt } from 'biggystring'
-import { asNumber, asObject, asOptional, asString } from 'cleaners'
+import {
+  asEither,
+  asNull,
+  asNumber,
+  asObject,
+  asOptional,
+  asString
+} from 'cleaners'
 import {
   EdgeCorePluginOptions,
   EdgeSpendInfo,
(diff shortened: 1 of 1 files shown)


src/swap/exolix.js
```diff
@@ -136,34 +147,33 @@ export function makeExolixPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
...
-
       // Swap the currencies if we need a reverse quote:
       const quoteParams =
         request.quoteFor === 'from'
           ? {
-              coin_from: safeFromCurrencyCode,
-              coin_to: safeToCurrencyCode,
-              deposit_amount: quoteAmount,
-              rate_type: 'fixed'
+              coinFrom: fromMainnetCode,
```
paullinator, line 154: You don't differentiate between the from and to amounts. If `request.quoteFor === 'to'` then the user is specifying the amount of crypto to receive and will be sending the `fromMainnetCode` currency. The code below just assumes from and to are in reverse. Major bug! Bug exists in current plugin which I'll have to fix for now to not allow reverse quotes.

paullinator (changes requested):

paullinator:
PR needs to be rebased on latest master (DO NOT merge master into your feature branch). Latest version already includes better use of cleaners and TS types.

src/swap/exolix.ts
```diff
@@ -140,7 +153,7 @@ export function makeExolixPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
       )
     }
 
-    const quoteAmount =
+    const nativeQuoteAmount =
```
paullinator, line 156: This is not a `native` amount. Should be renamed to `exchangeQuoteAmount` "exchange..." refers to denominations in the usual amount used by exchange. ie BTC, ETH not mETH, sats, bits.

src/swap/exolix.ts
```diff
@@ -151,41 +164,42 @@ export function makeExolixPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
...
-            coin_from: safeFromCurrencyCode,
-            coin_to: safeToCurrencyCode,
-            deposit_amount: quoteAmount,
-            rate_type: 'fixed'
+            coinFrom: fromCurrencyCode,
+            coinFromNetwork: fromMainnetCode,
+            coinTo: toCurrencyCode,
+            coinToNetwork: toMainnetCode,
+            amount: quoteAmount,
+            rateType: 'fixed'
```
paullinator, line 185: Both of these `quoteParams` are identical. You need to differentiate FROM vs TO quotes.

paullinator (changes requested):

src/swap/exolix.ts
devexolix: What do you mean? 
On line 156 we received `exchangeQuoteAmount` from `request.quoteFor = from|to` And if `request.quoteFor === 'to'` we will receive quoteAmount differen with quoteAmount as `request.quoteFor === 'from'`

src/swap/exolix.ts
devexolix: fixed

src/swap/exolix.ts
devexolix: @paullinator Do we need any more updates from our side?

devexolix (approved):

src/swap/exolix.ts
```diff
@@ -138,40 +141,60 @@ export function makeExolixPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
...
+      coinFromNetwork: fromMainnetCode,
+      coinTo: toCurrencyCode,
+      coinToNetwork: toMainnetCode,
+      amount: quoteAmount,
+      rateType: 'fixed'
+    }
+
+    // Set the withdrawal amount if we are quoting for the toCurrencyCode
+    if (request.quoteFor === 'to') {
+      quoteParams.withdrawalAmount = quoteAmount
```
Jon-edge, line 164: We leave the same `quoteAmount` in `amount` for a `to` quote?

ProcyonLotor123: Yes, `withdrawalAmount` has a priority over `amount` and we can leave it.

Jon-edge (approved):

src/swap/exolix.ts
```diff
@@ -173,7 +206,7 @@ export function makeExolixPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
       }
     } else {
       const nativeMin = await request.toWallet.denominationToNative(
-        rateResponse.min_amount,
+        rateResponse.minAmount.toString(),
```
paullinator, line 209: you already check minimums above. This logic now looks incorrect for 'to' quotes assuming minAmount is always the 'from' amount

paullinator (changes requested):

src/swap/exolix.ts
aguraan: @paullinator Thanks for noticing, corrected.

ProcyonLotor123: @paullinator Thanks for noticing, corrected.

paullinator (approved):
Please squash all into one commit before merge

PR #209, opened 2022-09-05
Add Thorchain support

paullinator:
---
  - https://app.asana.com/0/0/1201425242712343

### DIFF
--- a/src/swap/defi/thorchain.js
+++ b/src/swap/defi/thorchain.js
@@ -1,7 +1,14 @@
 // @flow
 
 import { div, lt, mul, sub, toFixed } from 'biggystring'
-import { asArray, asBoolean, asObject, asOptional, asString } from 'cleaners'
+import {
+  asArray,
+  asBoolean,
+  asNumber,
+  asObject,
+  asOptional,
+  asString
+} from 'cleaners'
 import {
   type EdgeCorePluginOptions,
   type EdgeCurrencyWallet,
@@ -20,7 +27,11 @@ import {
   isLikeKind,
   makeSwapPluginQuote
 } from '../../swap-helpers.js'
-import { fetchWaterfall } from '../../util/utils.js'
+import {
+  fetchInfo,
+  fetchWaterfall,
+  promiseWithTimeout
+} from '../../util/utils.js'
 
 const pluginId = 'thorchain'
 const swapInfo: EdgeSwapInfo = {
@@ -37,6 +48,7 @@ const THORNAME_DEFAULT = 'ej'
 const VOLATILITY_SPREAD_DEFAULT = 0.01
 const LIKE_KIND_VOLATILITY_SPREAD_DEFAULT = 0.0025
 const DO_CONSOLE_LOG = true
+const EXCHANGE_INFO_UPDATE_FREQ_MS = 60000
 const EVM_SEND_GAS = '80000'
 const EVM_TOKEN_SEND_GAS = '80000'
 const THOR_LIMIT_UNITS = '100000000'
@@ -92,8 +104,25 @@ const asPool = asObject({
   runeDepth: asString
 })
 
+const asExchangeInfo = asObject({
+  swap: asObject({
+    plugins: asObject({
+      thorchain: asObject({
+        volatilitySpread: asNumber,
+        likeKindVolatilitySpread: asNumber,
+        midgardServers: asArray(asString),
+        nineRealmsServers: asOptional(asArray(asString))
+      })
+    })
+  })
+})
+
 const asPools = asArray(asPool)
 type Pool = $Call<typeof asPool>
+type ExchangeInfo = $Call<typeof asExchangeInfo>
+
+let exchangeInfo: ExchangeInfo | void
+let exchangeInfoLastUpdate: number = 0
 
 export function makeThorchainPlugin(
   opts: EdgeCorePluginOptions
@@ -131,9 +160,9 @@ export function makeThorchainPlugin(
       } = request
       const likeKind = isLikeKind(fromCurrencyCode, toCurrencyCode)
 
-      const midgardServers: string[] = MIDGARD_SERVERS_DEFAULT
-      const likeKindVolatilitySpread: number = LIKE_KIND_VOLATILITY_SPREAD_DEFAULT
-      const volatilitySpread: number = VOLATILITY_SPREAD_DEFAULT
+      let midgardServers: string[] = MIDGARD_SERVERS_DEFAULT
+      let likeKindVolatilitySpread: number = LIKE_KIND_VOLATILITY_SPREAD_DEFAULT
+      let volatilitySpread: number = VOLATILITY_SPREAD_DEFAULT
 
       checkInvalidCodes(INVALID_CURRENCY_CODES, request, swapInfo)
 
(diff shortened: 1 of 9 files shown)


src/swap/defi/thorchain.js
```diff
@@ -0,0 +1,611 @@
...
+            sourcePool,
+            destPool,
+            volatilitySpreadFinal,
+            affiliateFee,
+            feeInDestCurrency
+          },
+          clog
+        )
+      } else {
+        throw new Error('TODO: TO quoting')
```
swansontec: Ooof, but I get it. Maybe add a more user-friendly string for the red box?

src/swap/defi/thorchain.js
```diff
@@ -0,0 +1,611 @@
...
+      let customNetworkFee
+      let customNetworkFeeKey
+
+      const customFeeTemplate = (fromWallet.currencyInfo.customFeeTemplate ??
+        [])[0]
+      const fromCurrencyInfo = fromWallet.currencyInfo
+      if (customFeeTemplate?.type === 'nativeAmount') {
+        customNetworkFee = gasRate
+        customNetworkFeeKey = customFeeTemplate.key
+      } else if (fromCurrencyInfo.defaultSettings?.customFeeSettings != null) {
```
swansontec, line 392: This code will almost certainly break in the future, since `defaultSettings` is an `any` dumping ground. I see that you are using the "official" way on the lines above, which tells me we probably need to get our currency plugins upgraded to use that instead.

src/swap/defi/thorchain.js
```diff
@@ -0,0 +1,611 @@
...
+    toCurrencyCode: string,
+    nativeAmount: string,
+    sourcePool: Pool,
+    destPool: Pool,
+    volatilitySpreadFinal: string,
+    affiliateFee: string,
+    feeInDestCurrency: string,
+    dontCheckLimits?: boolean
+  },
+  clog: Function
```
swansontec, line 564: `clog: EdgeLogMethod`. The `Function` type is pretty much an `any`.

src/swap/defi/thorchain.js
```diff
@@ -0,0 +1,611 @@
...
+    toNativeAmount,
+    toExchangeAmount,
+    limit
+  }
+}
+
+const buildSwapMemo = (params: BuildSwapMemoParams): string => {
+  const { chain, asset, address, limit, affiliateAddress, points } = params
+  // affiliate address could be a thorname, and the minimum received is not set in this example.
+  return `=:${chain}.${asset}:${address}:${limit}:${affiliateAddress}:${points}`
```
swansontec, line 775: This is an ASCII string, but we agreed to pass binary-data memos in hex. Where is the `base16.stringify(utf8.parse(...))`? I assume Thorchain wants the OP_RETURN or EVM extra data in UTF-8, which means we should make that conversion in the Thorchain plugin, not in the currency plugins.

Maybe we should have made the memo parameter an `Uint8Array`, just to be extra-explicit about these data-encoding things.

swansontec (changes requested):
This looks pretty solid overall.

src/swap/defi/thorchain.js
```diff
@@ -0,0 +1,611 @@
...
+  opts: EdgeCorePluginOptions
+): EdgeSwapPlugin {
+  const { initOptions, io, log } = opts
+  const { fetch } = io
+  const {
+    thorname = THORNAME_DEFAULT,
+    affiliateFeeBasis = AFFILIATE_FEE_BASIS_DEFAULT
+  } = initOptions
+  // eslint-disable-next-line no-console
+  const clog = (...args) => (DO_CONSOLE_LOG ? log(...args) : undefined)
```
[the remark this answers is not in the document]
paullinator: Actually, this was so that I could easily turn off logging completely without removing the code. But I'll make the change

src/swap/defi/thorchain.js
```diff
@@ -0,0 +1,611 @@
...
+        nativeAmount,
+        fromWallet,
+        toWallet,
+        quoteFor
+      } = request
+      const likeKind = isLikeKind(fromCurrencyCode, toCurrencyCode)
+
+      const midgardServers: string[] = MIDGARD_SERVERS_DEFAULT
+      const likeKindVolatilitySpread: number = LIKE_KIND_VOLATILITY_SPREAD_DEFAULT
+      const volatilitySpread: number = VOLATILITY_SPREAD_DEFAULT
```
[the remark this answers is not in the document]
paullinator: Yes this was mostly as attempt to break up my large PR into multiple commits easily. I could have removed this abstraction but then I would have to re-add it back later. I would have never had this had it not been for the info-server work in the later commit.

src/util/utils.js
```diff
@@ -0,0 +1,137 @@
...
+): Promise<T> => {
+  return new Promise((resolve, reject) => {
+    promise.then(v => resolve(v))
+    setTimeout(() => reject(new Error('PROMISE_TIMEOUT')), timeoutMs)
+  })
+}
+
+export function setFetch(fetchFunc: Function): void {
+  fetch = fetchFunc
+}
```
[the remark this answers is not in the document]
paullinator: 👍

src/swap/defi/thorchain.js
```diff
@@ -0,0 +1,611 @@
...
+
+      let responseJson = await iaResponse.json()
+      const inboundAddresses = asInboundAddresses(responseJson)
+
+      responseJson = await poolResponse.json()
+      const pools = asPools(responseJson)
+
+      // Check for supported chain and asset
+      const inAddressObject = inboundAddresses.find(
+        addrObj => addrObj.halted === false && addrObj.chain === fromMainnetCode
```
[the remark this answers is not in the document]
paullinator: Ok. I didn't realize false implies undefined as well. I intended for just an explicit boolean.

src/swap/defi/thorchain.js
paullinator: This is removed in a later commit when reverse quotes are added

src/swap/defi/thorchain.js
paullinator: I'm all for getting everything upgraded to the correct format, but I didn't want to scope creep this project so I adopted whatever models are currently in use. I won't make any changes to this PR for now.

src/swap/defi/thorchain.js
paullinator: Ok. fair enough

src/swap/defi/thorchain.js
paullinator: For the OP_RETURN we just pass an ascii string. That's the API for OP_RETURN as implemented. For EVM data, I convert the string to base16 here

https://github.com/EdgeApp/edge-exchange-plugins/pull/209/commits/6c74e38114a34392c2744befd3480ab91bfa0696#diff-d025ca361f3e8068e0bfdd23f1c3dcca99c3d2e5507cfa49c7503857d8e51f1eR434

src/swap/defi/thorchain.js
```diff
@@ -0,0 +1,611 @@
...
+const asPools = asArray(asPool)
+type Pool = $Call<typeof asPool>
+
+async function getAddress(
+  wallet: EdgeCurrencyWallet,
+  currencyCode: string
+): Promise<string> {
+  const addressInfo = await wallet.getReceiveAddress({ currencyCode })
+  return addressInfo.publicAddress
+}
```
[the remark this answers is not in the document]
paullinator: 👍

src/swap/defi/thorchain.js
```diff
@@ -0,0 +1,611 @@
...
+      if (!poolResponse.ok) {
+        const responseText = await poolResponse.text()
+        throw new Error(`Thorchain could not fetch pools: ${responseText}`)
+      }
+
+      let responseJson = await iaResponse.json()
+      const inboundAddresses = asInboundAddresses(responseJson)
+
+      responseJson = await poolResponse.json()
+      const pools = asPools(responseJson)
```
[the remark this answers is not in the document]
paullinator: done

paullinator:
Addressed comments but also added unit tests from a commit I stashed and forgot to include.

swansontec (approved):

PR #222, opened 2022-10-22
Add Swapuz

peachbits:
---
  - https://app.asana.com/0/0/1203186613849114

### DIFF
--- a/src/swap/swapuz.ts
+++ b/src/swap/swapuz.ts
@@ -1,0 +1,308 @@
+import { gt } from 'biggystring'
+import {
+  asDate,
+  asMaybe,
+  asNumber,
+  asObject,
+  asString,
+  Cleaner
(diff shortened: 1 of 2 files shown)


src/swap/swapuz.ts
```diff
@@ -0,0 +1,308 @@
...
+        return await getQuote('fix')
+      } catch (e) {
+        try {
+          return await getQuote('float')
+        } catch (e2) {
+          // Should throw the fixed-rate error
+          throw e
+        }
+      }
+    }
```
swansontec, line 250: This is fine, but since these are all promises, I wonder if `return await getQuote('fix').catch(() => getQuote('float')).catch(e => e // should throw fixed-rate error)` would be prettier to look at, taking advantage of the promise chaining instead of the try / catch syntax.

PR #223, opened 2022-10-24
Fix thorchain minimum quotes

paullinator:
---
  - https://app.asana.com/0/0/1203074693132557


swansontec (approved):
This all looks reasonable.

PR #229, opened 2022-10-31
Fix thorchain quotes

the-eridanus:


src/swap/defi/thorchain.ts
```diff
@@ -319,23 +320,23 @@ export function makeThorchainPlugin(
...
+      // const networkFee = calcNetworkFee(
+      //   toMainnetCode,
+      //   toCurrencyCode,
+      //   outboundGasRate,
+      //   pools
+      // )
 
       // Per Thorchain specs, user pays 2x the outbound transaction networkfee
-      let feeInDestCurrency = mul(networkFee, '2')
+      let feeInDestCurrency = outboundFee
```
paullinator: This appears to be missing the logic to convert the fee into the outbound asset for when the outbound asset is a token

paullinator (changes requested):

the-eridanus:
latest commit looks good 👍

paullinator (approved):

PR #230, opened 2022-11-03
Remove mutation of quote request

Jon-edge:
The same quote request object is reused between all exchange plugins when multiple quotes are being fetched.

Ensure no mutations of this request happen.


Jon-edge:
OTS approved by paul

PR #232, opened 2022-11-03
Lint cleanup

swansontec:
---
  - https://app.asana.com/0/0/1203287918505544

### DIFF
--- a/src/rate/bitmax.ts
+++ b/src/rate/bitmax.ts
@@ -35,7 +35,7 @@ export function makeBitMaxPlugin(opts: EdgeCorePluginOptions): EdgeRatePlugin {
             }
           ]
         } catch (e) {
-          log.warn(`Issue with Bitmax rate data structure ${e}`)
+          log.warn(`Issue with Bitmax rate data structure ${String(e)}`)
           break
         }
       }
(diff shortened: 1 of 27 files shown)


src/rate/bitmax.ts
```diff
@@ -35,7 +35,7 @@ export function makeBitMaxPlugin(opts: EdgeCorePluginOptions): EdgeRatePlugin {
             }
           ]
         } catch (e) {
-          log.warn(`Issue with Bitmax rate data structure ${e}`)
+          log.warn(`Issue with Bitmax rate data structure ${String(e)}`)
```
peachbits, line 38: We don't use the rate plugins anymore so we could delete them instead

src/rate/currencyconverterapi.ts
```diff
@@ -5,6 +5,10 @@ import {
   EdgeRatePlugin
 } from 'edge-core-js/types'
 
+const asInitOptions = asObject({
```
peachbits, line 8: Missed spookyswap and tombswap. Those plugins are buried real deep in the folder hierarchy for some reason.

peachbits (changes requested):

src/rate/currencyconverterapi.ts
swansontec: Added them.

src/rate/bitmax.ts
swansontec: I agree, but I think we need to do this in another PR.

peachbits (dismissed):

peachbits (approved):

PR #235, opened 2022-11-14
Add testing framework to run plugins in Node

paullinator:
---
  - https://app.asana.com/0/0/1203359209045189

### DIFF
--- a/src/util/matchJson.ts
+++ b/src/util/matchJson.ts
@@ -1,0 +1,47 @@
+/**
+ * Compares two JSON-like objects, returning false if they differ.
+ */
+export function matchJson(a: any, b: any): boolean {
+  // Use simple equality, unless a and b are proper objects:
+  if (
+    typeof a !== 'object' ||
+    typeof b !== 'object' ||
+    a == null ||
+    b == null
+  ) {
+    return a === b
+  }
+
+  // These must either be both arrays or both objects:
+  const aIsArray = Array.isArray(a)
+  const bIsArray = Array.isArray(b)
+  if (aIsArray !== bIsArray) return false
+
+  // Compare arrays in order:
+  if (aIsArray) {
+    if (a.length !== b.length) return false
+    for (let i = 0; i < a.length; ++i) {
+      if (!matchJson(a[i], b[i])) return false
+    }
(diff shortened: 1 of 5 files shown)


src/util/matchJson.ts
```diff
@@ -0,0 +1,39 @@
...
+  // Compare arrays in order:
+  if (aIsArray) {
+    if (a.length !== b.length) return false
+    for (let i = 0; i < a.length; ++i) {
+      if (!matchJson(a[i], b[i])) return false
+    }
+    return true
+  }
+  // These are both regular objects, so grab the keys:
+  const aKeys = Object.getOwnPropertyNames(a)
```
swansontec, line 28: This is the old corrupted copy from the GUI. Please grab the fresh version from edge-server-tools.

test/testpartner.ts
```diff
@@ -0,0 +1,99 @@
...
+} from 'edge-core-js'
+
+import edgeExchangePlugins from '../src'
+import { btcCurrencyInfo } from './fakeBtcInfo'
+import { makeFakePlugin } from './fakeCurrencyPlugin'
+import { ethCurrencyInfo } from './fakeEthInfo'
+
+async function main(): Promise<void> {
+  const allPlugins = {
+    bitcoin: makeFakePlugin(btcCurrencyInfo),
```
swansontec, line 15: This part is nice, injecting the currencyInfo like this.

swansontec (changes requested):
I would like to see some improvements like using the account snapshot feature, or turning this into a utility function so we can unit-test each of the different plugins, but those can be future projects. This is a fine starting point.

The only real issue is the outdated matchJson file you copied.

src/util/matchJson.ts
paullinator: Done. Updated w/fixup commit

test/testpartner.ts
paullinator: Thx

paullinator:
I agree we should use snapshots. I'll update the repo with the snapshot work in my Thorchain PR.

swansontec (dismissed):

PR #237, opened 2022-11-14
Audit and fix TO quotes to ensure precision accuracy

paullinator:
Ensure that the TO amount returned in the quote comes from the providers API and not just from the requested amount. This ensures precision accuracy


src/swap/foxExchange.ts
```diff
@@ -1,29 +1,11 @@
...
-  asBoolean,
-  asEither,
-  asNull,
-  asNumber,
-  asObject,
-  asOptional,
-  asString,
-  asUnknown
-} from 'cleaners'
 import {
```
Jon-edge, line 1: Do we need to disable these in the gui too?

Jon-edge (dismissed):
Approved. Do we need a task to update the GUI to also remove disabled swap plugins?

src/swap/foxExchange.ts
paullinator: eventually yes. But for now they will just never return a quote. Safe deprecation since our partners use these plugins in their app too.

PR #238, opened 2022-11-18
Thorchain DEX Aggregator

paullinator:


src/swap-helpers.ts
```diff
@@ -267,3 +278,23 @@ export const isLikeKind = (
...
+
+export const getTokenId = (
+  coreWallet: EdgeCurrencyWallet,
+  currencyCode: string
+): string | undefined => {
+  const { allTokens } = coreWallet.currencyConfig
+  return Object.keys(allTokens).find(
+    edgeToken => allTokens[edgeToken].currencyCode === currencyCode
+  )
+}
```
swansontec, line 290: We need to explicitly handle the `currencyCode === coreWallet.currencyInfo.currencyCode` case, since otherwise it would be possible to add a custom token named "ETH" on the Ethereum network, and have that break the main coin.

src/swap-helpers.ts
```diff
@@ -267,3 +278,23 @@ export const isLikeKind = (
...
+
+export const consify = (val: any): void =>
```
swansontec, line 292: This name means something else in other programming circles. Also, we should *never* be using console.log in here, unless it's temporary debugging code. So maybe call it `prettyDebugLog` or something.

test/testpartner.ts
```diff
@@ -1,59 +1,119 @@
...
 import edgeExchangePlugins from '../src'
+import { avaxCurrencyInfo } from './fakeAvaxInfo'
 import { btcCurrencyInfo } from './fakeBtcInfo'
 import { makeFakePlugin } from './fakeCurrencyPlugin'
 import { ethCurrencyInfo } from './fakeEthInfo'
 
+const DUMP_USER = false
+const DUMP_USER_FILE = './test/fakeUserDump.json'
+
+const asDateStr = (raw: string): Date => new Date(raw)
```
swansontec, line 20: This is not a cleaner. Cleaners only take `any`, not `string`. I think you want `asDate`, which also handles the string conversion like you have here.

test/testpartner.ts
```diff
@@ -1,59 +1,119 @@
...
+    await avaxWallet.enableTokens(avaxEnabledTokens)
+
+    const data = await world.dumpFakeUser(account)
+    const dump = {
+      loginKey: account.loginKey,
+      data
+    }
+    fs.writeFileSync(DUMP_USER_FILE, JSON.stringify(dump), { encoding: 'utf8' })
+    process.exit(0)
+  }
```
swansontec, line 99: Since setting the boolean completely changes what the program does (building a fake user vs. testing a plugin), should there really be two separate script files, one to test plugins and another to re-build the fake user?

src/swap/defi/thorchainDa.ts
```diff
@@ -0,0 +1,547 @@
...
+  ethereum: '0xf892fef9da200d9e84c9b0647ecff0f34633abe8',
+  avalanche: '0x69ba883af416ff5501d54d5e27a1f497fbd97156'
+}
+
+export function makeThorchainDaPlugin(
+  opts: EdgeCorePluginOptions
+): EdgeSwapPlugin {
+  const { io, log } = opts
+  const { fetch } = io
+  // const fetch = nodeFetch ?? io.fetch
```
swansontec, line 112: Leftover line.

src/swap-helpers.ts
```diff
@@ -267,3 +278,23 @@ export const isLikeKind = (
...
+  console.log(JSON.stringify(val, null, 2))
+
+export const makeQueryParams = (params: JsonObject): string => {
+  const cleaned = asQueryParams(params)
+  return Object.entries(cleaned)
+    .map(([key, value]) => `${key}=${value.toString()}`)
+    .join('&')
+}
```
swansontec, line 300: We should not be cleaning stuff our own program produces. Cleaners are for "outside" data sources like disk or network. A better solution is to delete the cleaner and actually type what we mean:

```ts
export interface QueryParams {
  [key: string]: number | string | boolean | undefined | null
}
```

You should also handle the `null` case by skipping the `=${value}` bit, since sometimes you can put raw keys into query strings like `?count=1&sorted&descending`. Maybe `.map(([key, value]) => value == null ? key : '${key}=${encodeURIComponent(value)}')`, to also handle escaping.

It's also worth noting that `Object.entries` is not going to work on older phones, so those aren't going to support DEX swaps on Thorchain (TypeError: Object.entries is not a function). You could use `Object.keys`, which is much older, and then grab out the value using [], or we could just be fine with the failed quotes.

swansontec (changes requested):

src/swap-helpers.ts
```diff
@@ -69,6 +76,10 @@ export function makeSwapPluginQuote(
         await fromWallet.saveTx(broadcastedTransaction)
       }
       tx.metadata = { ...(opts?.metadata ?? {}), ...tx.metadata }
+      if (metadataNotes != null) {
+        tx.metadata.notes = `${metadataNotes}\n\n` + (tx.metadata.notes ?? '')
```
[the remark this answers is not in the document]
paullinator: Seemed harmless and make the code simpler.

src/swap/defi/thorchainDa.ts
paullinator: io.fetch is broken for some API endpoints when running in node. This was added as a convenience so unit tests could work. Either need to troubleshoot io.fetch in node or leave this is for testing convenience (and as a note).

src/swap-helpers.ts
paullinator: done

src/swap-helpers.ts
paullinator: this is for temporary debugging. I renamed to `prettyLogObject` since it's specifically for logging objects.

src/swap-helpers.ts
paullinator: Problem is typescript doesn't like a fixed object passed into a function that accepts a map. This is the object being created

```
      const quoteParams: ThorSwapQuoteParams = {
        sellAsset:
          `${fromMainnetCode}.${fromCurrencyCode}` + (fromTokenId ?? ''),
        buyAsset: `${toMainnetCode}.${toCurrencyCode}` + (toTokenId ?? ''),
        sellAmount,
        slippage: (volatilitySpreadFinal * 100).toString(),
        recipientAddress: toAddress,
        senderAddress: fromAddress,
        affiliateAddress: thorname,
        affiliateBasisPoints: affiliateFeeBasis
      }
      ```
      
      Once this is passed into `makeQueryParams` which accepts a map, TS complains that
      
      ```
      Argument of type 'ThorSwapQuoteParams' is not assignable to parameter of type 'QueryParams'.
  Index signature for type 'string' is missing in type 'ThorSwapQuoteParams'.ts(2345)
```

src/swap/defi/thorchainDa.ts
paullinator: Example error I get with `io.fetch` under node that works with `node-fetch`

`"{\"status_code\":1,\"message\":\"Unknown API endpoint /v2/thorchain/inbound_addresses\"}"`

src/swap-helpers.ts
paullinator: Sam figured out a hack to get this to work as a type instead of cleaner. Basically `ThorSwapQuoteParams` needs to be a `type` not an `interface` then it can be cast into a map. I updated the fixup commit with this change

test/testpartner.ts
paullinator: done

test/testpartner.ts
paullinator: done

src/swap-helpers.ts
paullinator: renamed to prettyObjectLog and moved to utils.ts

src/swap-helpers.ts
paullinator: ok. good catch.

swansontec (approved):

PR #246, opened 2022-12-09
Paul/tc fix spread

paullinator:
---
  - https://app.asana.com/0/0/1203511199083806

### DIFF
--- a/src/swap/defi/thorchain.ts
+++ b/src/swap/defi/thorchain.ts
@@ -27,7 +27,12 @@ import {
   isLikeKind,
   makeSwapPluginQuote
 } from '../../swap-helpers'
-import { fetchInfo, fetchWaterfall, promiseWithTimeout } from '../../util/utils'
+import {
+  convertRequest,
+  fetchInfo,
+  fetchWaterfall,
+  promiseWithTimeout
+} from '../../util/utils'
 import abi from './abi/THORCHAIN_SWAP_ABI'
 import erc20Abi from './abi/UNISWAP_V2_ERC20_ABI'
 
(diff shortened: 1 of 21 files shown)


src/swap/defi/thorchain.ts
```diff
@@ -961,3 +1027,57 @@ const convertChainAmountToAsset = (
...
+
+  if (volatilitySpreadFinal == null) {
+    const likeKind = isLikeKind(fromCurrencyCode, toCurrencyCode)
+
+    volatilitySpreadFinal = likeKind
+      ? likeKindVolatilitySpread
+      : volatilitySpread
+  }
+
+  log.warn(`getVolatilitySpread: ${volatilitySpreadFinal.toString()}`)
```
swansontec, line 1081: This should be info, not a warning. Nothing has gone wrong, and verbose logging can activate it if needed. It should also be moved out of this routine into the `fetchSwapQuote` body, since unit tests should *never* write to the console.

Yes, we don't enforce this very well, but a test's only job is to pass or fail, not produce output. If you're trying to debug a failing test, and some irrelevant neighbor is being noisy, that's not helpful.

swansontec (changes requested):
The logic looks good, but the logging is too aggressive.

src/swap/defi/thorchain.ts
paullinator: Happy to move it out to fetchSwapQuote, but I used log.warn because I want to make sure we get this value in the logs if a swap ends up failing due to spread. This is a debug value that only helps after the fact. If user runs into a failed swap, turning on verbose wont help as by then it's too late.

swansontec: Makes sense. Approved.

swansontec (approved):

PR #252, opened 2023-01-03
Send swapPluginType in EdgeSwapInfo and EdgeSwapQuote

paullinator:
---
  - https://app.asana.com/0/0/1203632680642482

### DIFF
--- a/src/swap-helpers.ts
+++ b/src/swap-helpers.ts
@@ -32,6 +32,7 @@ export function ensureInFuture(
 
 export function makeSwapPluginQuote(
   request: EdgeSwapRequest,
+  swapInfo: EdgeSwapInfo,
   fromNativeAmount: string,
   toNativeAmount: string,
   tx: EdgeTransaction,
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -100,6 +100,7 @@
     "stream-browserify": "^3.0.0",
     "sucrase": "^3.9.5",
     "typescript": "^4.1.2",
+    "updot": "^1.1.7",
```
swansontec, line 103: We discussed this before, and `updot` should only be a dependency for the GUI, where `yarn link` doesn't work. If you prefer the `updot` workflow in general, then it's up to you to install it globally on your computer, such as with with `npm i -g updot` so you can use it wherever you like. Otherwise, the normal NPM / yarn symlink workflows work fine in these other places.

swansontec (changes requested):
The `isDex` changes look great, but the out-of-scope updot thing isn't OK.

package.json
paullinator: ok. thats fair. I rebased the PR to exclude that commit.

swansontec (approved):

PR #254, opened 2023-01-04
Add changelly

paullinator:
---
  - https://app.asana.com/0/0/1203629760009427


peachbits:
Replaced by https://github.com/EdgeApp/edge-exchange-plugins/pull/471

PR #257, opened 2023-01-05
Add max swap

peachbits:
---
  - https://app.asana.com/0/0/1200828710714395


src/swap-helpers.ts
```diff
@@ -118,11 +125,16 @@ export async function makeSwapPluginQuote(
   return out
 }
 
+interface getMaxSwappableRequest {
```
swansontec, line 128: Type names must begin with uppercase.

src/swap-helpers.ts
```diff
@@ -38,12 +39,14 @@ export interface SwapOrder {
   expirationDate?: Date
   preTx?: EdgeTransaction
   metadataNotes?: string
+  customNetworkFee?: JsonObject
```
swansontec, line 42: We should not need to pass this in again, since the `spendInfo` already has it. It should be up to the `fetchSwapPluginInner` methods to preserve the fees and stick them into the `spendInfo` as needed.

src/swap/changehero.ts
```diff
@@ -292,15 +293,16 @@ export function makeChangeHeroPlugin(
       expirationDate: new Date(Date.now() + expirationFixedMs)
     }
 
-    return await makeSwapPluginQuote(order)
+    return order
```
swansontec, line 297: Please go through all these methods and replace the `const order = { ... }; return order` logic with just `return { ... }`.

Adding the extra `order` variable means TypeScript has less information to do inference with, and this will cause problems later. If you do `return { ... }`, though, TypeScript knows that this has to be a `SwapOrder` structure (because the function type says so), and this clue will let it do better inference / give better error messages.

swansontec (changes requested):
The logic here seems pretty solid, and I am happy with the way it came out. Most of the changes I have are minor, but repeated across many files.

This was a funny PR to review, because I would leave comments on one commit, only to find the next commit completely changing things so the comments weren't needed. Lots of big seismic shifts in this one.

src/swap/changenow.ts
```diff
@@ -171,7 +172,9 @@ export function makeChangeNowPlugin(
 
       async function swapSell(
         flow: 'fixed-rate' | 'standard'
-      ): Promise<EdgeSwapQuote> {
+      ): Promise<SwapOrder> {
```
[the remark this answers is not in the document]
peachbits: Agreed it was dangerous to keep them in scope. I redid it so the inner function is outside of fetchSwapQuote

src/swap-helpers.ts
peachbits: interface no longer exists after addressing changes

src/swap/transfer.ts
```diff
@@ -44,29 +43,22 @@ export function makeTransferPlugin(
...
+      const spendInfo = {
         currencyCode: request.fromCurrencyCode,
         spendTargets: [
           {
             nativeAmount: request.nativeAmount,
             publicAddress: toAddress
           }
         ]
-      })
+      }
```
swansontec: In the previous version of this PR, you added a bunch of swap parameters to this `spendInfo`. Those changes no longer seem to exist, so we need to get them back.

src/swap/defi/uni-v2-based/plugins/spookySwap.ts
```diff
@@ -41,6 +41,10 @@ export function makeSpookySwapPlugin(
   const { quiknodeApiKey } = asInitOptions(opts.initOptions)
   const provider = getFtmProvider(quiknodeApiKey)
 
+  // For max swaps, save the gas price from the first getSwapTransactions call and pass it into the second
+  let isMaxSwap = false
+  let previousGasPrice: string | undefined
```
swansontec, line 46: Using these external variables could cause a race condition if somebody requests multiple swaps in parallel. I would like to see the previous version of this PR, so we can kit-bash those changes with these ones to get a single stable solution.

swansontec (changes requested):

src/swap/changenow.ts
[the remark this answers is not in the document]
swansontec: This looks much, much prettier. Thanks.

src/swap/transfer.ts
swansontec: Found it:

```ts
          swapData: {
            isEstimate: false,
            plugin: { ...swapInfo },
            payoutAddress: toAddress,
            payoutCurrencyCode: requestInner.fromCurrencyCode,
            payoutNativeAmount: requestInner.nativeAmount, // Wrong
            payoutWalletId: requestInner.toWallet.id
          }
```

src/swap/defi/uni-v2-based/plugins/spookySwap.ts
peachbits: The previous solution was to pipe the custom network fee from the first makeSpend (in getMaxSwappable) and return it with the newRequest back to the swap plugin. Then it would be passed it into the inner function with the request and other args. 

That solution is better and would be a step towards resolving another race condition where calling two makeSpends, without custom fees, can result in different fees for each one. We don't currently have a way to grab the fee from a makeSpend (feeRateUsed is for display and doesn't map perfectly for all currency plugins) but if we did we could grab it make sure it's included in the second go around with the max amount. 

For the solution that's currently in the PR, I could make utility function that holds a uid for each request that the plugins can access. I'll add that as a fixup

src/swap-helpers.ts
```diff
@@ -122,6 +122,42 @@ export async function makeSwapPluginQuote(
   return out
 }
 
+export const getMaxSwappable = async (
```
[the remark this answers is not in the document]
peachbits: done

src/swap/transfer.ts
peachbits: yup missed that, done

src/swap/defi/uni-v2-based/plugins/spookySwap.ts
peachbits: Added a fixup that resolves the race condition using a map at a higher scope

swansontec (approved):

PR #261, opened 2023-02-07
Fix thorchain

paullinator:
---
  - https://app.asana.com/0/0/1203879199448851

### DIFF
--- a/src/swap/defi/thorchain.ts
+++ b/src/swap/defi/thorchain.ts
@@ -53,6 +53,7 @@ const asInitOptions = asObject({
 })
 
 export const MIDGARD_SERVERS_DEFAULT = ['https://midgard.thorchain.info']
+export const THORNODE_SERVERS_DEFAULT = ['https://thornode.ninerealms.com']
 export const EXPIRATION_MS = 1000 * 60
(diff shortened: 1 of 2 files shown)


src/swap/defi/thorchain.ts
```diff
@@ -213,11 +214,14 @@ export function makeThorchainPlugin(
...
-  const { thorname, affiliateFeeBasis } = asInitOptions(opts.initOptions)
+  const { thorname, affiliateFeeBasis, ninerealmsClientId } = asInitOptions(
+    opts.initOptions
+  )
   const affiliateFee = div(affiliateFeeBasis, '10000', DIVIDE_PRECISION)
 
   const headers = {
-    'Content-Type': 'application/json'
+    'Content-Type': 'application/json',
+    'x-client-id': ninerealmsClientId
```
swansontec, line 224: These headers are being sent to both thorswap and midgard servers. Do they both care about `ninerealmsClientId`?

swansontec (approved):
Approved with one question.

src/swap/defi/thorchain.ts
paullinator: They don't but thorswap just ignores it.

PR #262, opened 2023-02-15
Add Optimism

peachbits:
---
  - https://app.asana.com/0/0/1203970651584029

### DIFF
--- a/src/swap-helpers.ts
+++ b/src/swap-helpers.ts
@@ -301,6 +301,10 @@ const defaultCurrencyCodeTranscriptionMap: CurrencyCodeTranscriptionMap = {
   }
 }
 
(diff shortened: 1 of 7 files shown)


src/swap-helpers.ts
```diff
@@ -301,6 +301,10 @@ const defaultCurrencyCodeTranscriptionMap: CurrencyCodeTranscriptionMap = {
   }
 }
 
+const defaultMainnetTranscriptionMap: MainnetPluginIdTranscriptionMap = {
+  optimism: 'OP' // mainnet code it ETH
```
swansontec, line 305: It comment weird?

swansontec (approved):
Approve, but think one fix maybe?

PR #263, opened 2023-02-16
Implement Lifi DEX aggregator

paullinator:
---
  - https://app.asana.com/0/0/1203860324834591

### DIFF
--- a/src/swap/defi/lifi.ts
+++ b/src/swap/defi/lifi.ts
@@ -1,0 +1,383 @@
+import { round } from 'biggystring'
+import { asArray, asNumber, asObject, asOptional, asString } from 'cleaners'
+import {
+  EdgeCorePluginOptions,
+  EdgeSpendInfo,
+  EdgeSwapInfo,
+  EdgeSwapPlugin,
+  EdgeSwapQuote,
+  EdgeSwapRequest,
+  EdgeTransaction,
+  SwapCurrencyError
+} from 'edge-core-js/types'
+
+import {
+  checkInvalidCodes,
+  getMaxSwappable,
+  InvalidCurrencyCodes,
+  makeSwapPluginQuote,
+  SwapOrder
+} from '../../swap-helpers'
+import { div18 } from '../../util/biggystringplus'
+import {
+  convertRequest,
+  fetchInfo,
+  fetchWaterfall,
+  getAddress,
+  hexToDecimal,
+  makeQueryParams,
+  promiseWithTimeout
+} from '../../util/utils'
+import { asNumberString, EdgeSwapRequestPlugin, StringMap } from '../types'
+import { getEvmApprovalData } from './defiUtils'
+
+const pluginId = 'lifi'
+const swapInfo: EdgeSwapInfo = {
+  pluginId,
+  isDex: true,
+  displayName: 'LI.FI',
+  supportEmail: 'support@edge.app'
+}
+
+const asInitOptions = asObject({
+  affiliateFeeBasis: asOptional(asString, '50'),
+  appId: asOptional(asString, 'edge'),
+  integrator: asOptional(asString, 'edgeapp')
+})
+
+const LIFI_SERVERS_DEFAULT = ['https://li.quest']
+const EXPIRATION_MS = 1000 * 60
+const EXCHANGE_INFO_UPDATE_FREQ_MS = 60000
+
+export const INVALID_CURRENCY_CODES: InvalidCurrencyCodes = {
+  from: {},
+  to: {
+    zcash: ['ZEC']
+  }
+}
+
+// Network names that don't match parent network currency code
+const MAINNET_CODE_TRANSCRIPTION: StringMap = {
+  arbitrum: 'ARB',
+  aurora: 'AUR',
+  avalanche: 'AVA',
+  binancesmartchain: 'BSC',
+  celo: 'CEL',
+  cronos: 'CRO',
+  ethereum: 'ETH',
+  evmos: 'EVM',
+  fantom: 'FTM',
+  fuse: 'FUS',
+  gnosis: 'DAI',
+  harmony: 'ONE',
+  moonbeam: 'MOO',
+  moonriver: 'MOR',
+  okexchain: 'OKT',
+  optimism: 'OPT',
+  polygon: 'POL',
+  velas: 'VEL'
+}
+
+const asExchangeInfo = asObject({
+  swap: asObject({
+    plugins: asObject({
+      lifi: asOptional(
+        asObject({
+          // perAssetSpread: asOptional(asArray(asAssetSpread)),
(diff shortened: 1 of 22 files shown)


src/swap/defi/lifi.ts
```diff
@@ -0,0 +1,392 @@
...
+      toContractAddress = toCurrencyCode
+    } else {
+      toContractAddress = toToken?.networkLocation?.contractAddress
+    }
+
+    if (fromContractAddress == null || toContractAddress == null) {
+      throw new SwapCurrencyError(swapInfo, fromCurrencyCode, toCurrencyCode)
+    }
+
+    const { appId, integrator } = asInitOptions(opts.initOptions)
```
peachbits, line 218: Nit pick, these consts and the lifiServers declaration should be after all of the early return scenarios (next to the now declaration)

src/swap/defi/lifi.ts
```diff
@@ -0,0 +1,392 @@
...
+        const exchangeInfoResponse = await promiseWithTimeout(
+          fetchInfo(fetch, `v1/exchangeInfo/${appId}`)
+        )
+
+        if (exchangeInfoResponse.ok === true) {
+          exchangeInfo = asExchangeInfo(await exchangeInfoResponse.json())
+          exchangeInfoLastUpdate = now
+        } else {
+          // Error is ok. We just use defaults
+          log('Error getting info server exchangeInfo. Using defaults...')
```
peachbits, line 262: Would change this to warn and print the quoteResponse.text()

src/swap/defi/lifi.ts
```diff
@@ -0,0 +1,392 @@
...
+        }
+      } catch (e: any) {
+        log(
```
peachbits, line 265: change to log.warn

src/swap/defi/lifi.ts
```diff
@@ -0,0 +1,392 @@
...
+          // likeKindVolatilitySpread: asOptional(asNumber),
+          // daVolatilitySpread: asOptional(asNumber),
+          lifiServers: asOptional(asArray(asString))
+        })
+      )
+    })
+  })
+})
+
+// const asToken = asObject({
```
peachbits, line 107: Are these here because they might be relevant later?

src/swap/defi/lifi.ts
```diff
@@ -0,0 +1,392 @@
...
+        log(
+          'Error getting info server exchangeInfo. Using defaults...',
+          e.message
+        )
+      }
+    }
+
+    if (exchangeInfo != null) {
+      const { lifi } = exchangeInfo.swap.plugins
+      lifiServers = lifi?.lifiServers ?? lifiServers
```
peachbits, line 274: Another nit pick: this block can be removed if you update the lifiServers array where you set 'exchangeInfoLastUpdate = now' in the try/catch

peachbits (changes requested):
Mostly nit-picky style comments

test/testpartner.ts
```diff
@@ -63,16 +65,14 @@ async function main(): Promise<void> {
...
-        secret: ''
-      },
-      letsexchange: {
-        apiKey: '',
-        affiliateId: ''
-      }
+      lifi: true,
+      polygon: true,
+      piratechain: true
+      // thorchainda: true,
```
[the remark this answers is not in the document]
paullinator: Doesn't matter. `testpartner.ts` is not run with any automated testing. It is a utility script for developing plugins in Node vs having to debug the webview.

test/fakeEthInfo.ts
```diff
@@ -4,7 +4,7 @@ import { FakeSettings } from './fakeCurrencyPlugin'
 
 const defaultSettings: FakeSettings = {
   customFeeSettings: ['gasLimit', 'gasPrice'],
-  publicAddress: '0x0d73358506663d484945ba85d0cd435ad610b0a0',
+  publicAddress: '0x6f8bc2d82175489f50537d868bcf2e4f15c3fa95',
```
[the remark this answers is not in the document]
paullinator: True, but also not very relevant. These are only used for the testpartner.ts utility development script. This was changed to match real public addresses that have balances in them in case the Lifi API was checking. I'll revert back.

src/swap/defi/lifi.ts
```diff
@@ -0,0 +1,392 @@
...
+  promiseWithTimeout
+} from '../../util/utils'
+import { asNumberString, EdgeSwapRequestPlugin, StringMap } from '../types'
+import { getEvmApprovalData } from './defiUtils'
+
+const pluginId = 'lifi'
+const swapInfo: EdgeSwapInfo = {
+  pluginId,
+  isDex: true,
+  displayName: 'Li.Fi',
```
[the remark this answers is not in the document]
paullinator: ok

src/swap/defi/lifi.ts
```diff
@@ -0,0 +1,392 @@
...
+  EdgeSwapInfo,
+  EdgeSwapPlugin,
+  EdgeSwapQuote,
+  EdgeSwapRequest,
+  EdgeTransaction,
+  SwapCurrencyError
+} from 'edge-core-js/types'
+
+// XXX hack. io.fetch is broken in node for some APIs
+// import nodeFetch from 'node-fetch'
```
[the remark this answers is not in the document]
paullinator: This is a useful comment as io.fetch is broken under node. I frequently uncomment this out for when plugins are run by testpartner.ts

src/swap/defi/thorchainDa.ts
```diff
@@ -127,9 +122,12 @@ export function makeThorchainDaPlugin(
   const { io, log } = opts
   const { fetch } = io
   // const fetch = nodeFetch ?? io.fetch
```
swansontec, line 123: I agree with Matthew that all this leftover `fetch` stuff needs to go. *Especially* since the core just directly exposes `node-fetch` as its `io.fetch` function. If you are saying `io.fetch` doesn't work, then `import 'node-fetch'` shouldn't work any better, because it's the exact same thing. Perhaps the core's version is outdated or something? Whatever the problem is, it should be a one-line fix in the core, not a bunch of leftover junk in the plugins.

swansontec (changes requested):

swansontec (approved):
Per discussion, this is approved because we will debug the core in a separate task.

src/swap/defi/lifi.ts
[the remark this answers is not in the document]
swansontec: I went ahead and fixed it in the core, then added a commit to the end of this PR to make the requested changes.

src/swap/defi/thorchainDa.ts
swansontec: I went ahead and fixed it in the core, then added a commit to the end of this PR to make the requested changes.

PR #269, opened 2023-04-19
Add default zkSync "ZKSYNC" mainnet code

samholmes:
Some plugins don't support zkSync yet so we don't know what the 
mainnet would be called. 
This provides protection against a plugin from adding supporting and 
defaulting to the ETH mainnet code.

### CHANGELOG

- Changed: Transcribe zkSync mainnet code to ZKSYNC

### Dependencies

 none

### Description

 none

### DIFF
--- a/src/swap-helpers.ts
+++ b/src/swap-helpers.ts
@@ -302,7 +302,8 @@ const defaultCurrencyCodeTranscriptionMap: CurrencyCodeTranscriptionMap = {
 }
 
 const defaultMainnetTranscriptionMap: MainnetPluginIdTranscriptionMap = {
-  optimism: 'OP' // mainnet code is ETH
+  optimism: 'OP', // mainnet code is ETH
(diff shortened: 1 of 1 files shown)


src/swap/changehero.ts
```diff
@@ -43,6 +43,7 @@ const asInitOptions = asObject({
   apiKey: asString
 })
 
+// See https://changenow.io/currencies for list of supported currencies
```
swansontec: This file is for ChangeHero, but the link is for ChangeNow?

swansontec: Oh, I see the very next commit fixes it. The next commit should be a fixup / squash  then.

swansontec (approved):
I think you should squash all of the comment-only commits together, but that's just a taste thing. The actual changes are fine.

src/swap/changehero.ts
samholmes: I squashed all the URI changes to one commit.

PR #274, opened 2023-05-10
Prevent tc swaps that would cause negative dest amounts

paullinator:
### CHANGELOG

 none

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1204553834059679

### DIFF
--- a/src/swap/defi/thorchain.ts
+++ b/src/swap/defi/thorchain.ts
@@ -1,4 +1,4 @@
-import { add, div, gt, lt, mul, sub, toFixed } from 'biggystring'
+import { add, div, gt, lt, lte, mul, sub, toFixed } from 'biggystring'
 import {
   asArray,
   asBoolean,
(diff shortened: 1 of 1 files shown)


src/swap/defi/thorchain.ts
```diff
@@ -438,6 +438,10 @@ export function makeThorchainPlugin(
     }
     const { fromNativeAmount, toNativeAmount, limit } = calcResponse
 
+    if (lte(limit, '0')) {
+      throw new Error('Swap would produce negative return')
```
swansontec, line 442: Returning a "below limit" error would be a better UX, but I have no idea how to calculate the parameters for that error, so this is fine. This will be the lowest priority message, since it isn't one of the known types, but at least we don't give a bogus quote.

swansontec (approved):

src/swap/defi/thorchain.ts
paullinator: Yes, that's what I wanted, but I have no way of calculating it either.

PR #281, opened 2023-08-14
Use the `value` from the LI.FI quote response data directly

samholmes:
The mainnet value must be sent as it is calculated by the LI.FI API because it includes mainnet amount needed to pay for bridge fees within the smart contract.

### CHANGELOG

- Fixed: LI.FI on-chain transactions no longer revert due to missing bridge fees

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1204953687205427

### DIFF
--- a/src/swap/defi/lifi.ts
+++ b/src/swap/defi/lifi.ts
@@ -291,10 +291,10 @@ export function makeLifiPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
 
     const quoteJson = await quoteResponse.json()
     const quote = asV1Quote(quoteJson)
-    const { estimate, includedSteps } = quote
+    const { estimate, includedSteps, transactionRequest } = quote
     const { approvalAddress, toAmountMin } = estimate
 
-    const { data, gasLimit, gasPrice } = quote.transactionRequest
+    const { data, gasLimit, gasPrice } = transactionRequest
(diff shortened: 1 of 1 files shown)


src/swap/defi/lifi.ts
```diff
@@ -312,7 +312,7 @@ export function makeLifiPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
         spendTargets: [
           {
             memo: approvalData,
-            nativeAmount: '0',
+            nativeAmount: mul(transactionRequest.value, '1'),
```
paullinator, line 315: Yikes. This is the approval transaction and should never send actual funds. This could cause a loss of funds for the user.

src/swap/defi/lifi.ts
```diff
@@ -333,7 +333,7 @@ export function makeLifiPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
       spendTargets: [
         {
           memo: data,
-          nativeAmount: nativeAmount,
+          nativeAmount: mul(transactionRequest.value, '1'),
```
paullinator, line 336: Why would this ever be different from `nativeAmount`. Lifi only supports `from` quotes, so if the user specifies that they only want to send 0.1234 of an asset, we can't send more than that.

paullinator (changes requested):

paullinator (approved):

PR #284, opened 2023-08-29
Add TC streaming spreads

paullinator:
### CHANGELOG

added: Thorchain streaming spreads

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205371018874225

### DIFF
--- a/src/swap/defi/thorchain.ts
+++ b/src/swap/defi/thorchain.ts
@@ -79,6 +79,8 @@ const STREAMING_QUANTITY_NOSTREAM = 1
 // ----------------------------------------------------------------------------
 export const VOLATILITY_SPREAD_DEFAULT = 0.0075
 export const LIKE_KIND_VOLATILITY_SPREAD_DEFAULT = 0.005
+export const VOLATILITY_SPREAD_STREAMING_DEFAULT = 0.001
+export const LIKE_KIND_VOLATILITY_SPREAD_STREAMING_DEFAULT = 0
 export const PER_ASSET_SPREAD_DEFAULT: AssetSpread[] = [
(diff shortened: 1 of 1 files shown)


src/swap/defi/thorchain.ts
```diff
@@ -195,9 +195,12 @@ export const asExchangeInfo = asObject({
...
       thorchain: asObject({
-        perAssetSpread: asArray(asAssetSpread),
-        volatilitySpread: asNumber,
-        likeKindVolatilitySpread: asNumber,
+        // perAssetSpread: asArray(asAssetSpread),
+        perAssetSpreadStreaming: asOptional(asArray(asAssetSpread)),
+        // volatilitySpread: asNumber,
+        volatilitySpreadStreaming: asOptional(asNumber),
+        // likeKindVolatilitySpread: asNumber,
+        likeKindVolatilitySpreadStreaming: asOptional(asNumber),
```
samholmes, line 203: Instead of keeping comments, we can just rely on our repository history for access to the code prior to the change. Unless these are still valid for the info server; then why not leave them with a deprecated comment?

samholmes (approved):
Approved with minor nit. I take it these new spreads are based on the fact that some new streaming version of the TC API is being used?

src/swap/defi/thorchain.ts
paullinator: Crud. I just realize we still need to apply the non-streaming spreads in case we get a non-streaming quote from the api. another change incoming.

paullinator:
Needs another review @samholmes as the PR was rewritten

PR #291, opened 2023-09-21
Remove EdgeMetaToken usage

peachbits:
---
  - https://app.asana.com/0/0/1205544705952481

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- changed: Update getInOutTokenAddresses to use EdgeTokens so it can be used for any token
 - changed: Block VELO trading from providers that rely on `currencyCode`
 - changed: Add `fantom` pluginId check to spookySwap and tombSwap
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- changed: Block VELO trading from providers that rely on `currencyCode`
 - changed: Add `fantom` pluginId check to spookySwap and tombSwap
 
 ## 0.21.6 (2023-09-14)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # edge-exchange-plugins
 
+## Unreleased
+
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- changed: Update getInOutTokenAddresses to use EdgeTokens
```
swansontec, line 5: I don't think we need changelog entries for every commit, especially with weird internal details. From a user-facing standpoint, what changed? I would expect:

- fixed: Block VELO tokens from currency-code based swaps, since there are two tokens with the same name.
- fixed: Lock spookySwap and tombSwap to only work on the Fantom chain.

src/swap/defi/defiUtils.ts
```diff
@@ -34,29 +19,36 @@ export interface InOutTokenAddresses {
...
+  const fromToken: EdgeToken | undefined = isFromNativeCurrency
+    ? undefined
+    : allTokens[fromTokenId]
+  const toToken: EdgeToken | undefined = isToNativeCurrency
+    ? undefined
+    : allTokens[toTokenId]
+
+  const fromTokenAddress: string =
+    fromToken?.networkLocation?.contractAddress ?? ''
+  const toTokenAddress: string = toToken?.networkLocation?.contractAddress ?? ''
```
swansontec, line 41: Since the `networkLocation` parameter is untyped, we should really use a cleaner:

```js
const asContractLocation = asObject({
  contractAddress: asString
})

const fromTokenAddress = asMaybeContractLocation(toToken?.networkLocation)?.contractAddress ?? ''
```

Other chains have a different structure, so this make the assumption explicit. Ripple tokens are totally weird, for example, and will probably need a different cleaner.

swansontec (changes requested):

CHANGELOG.md
peachbits: The change is these evm swap providers now work with any token regardless of currency code. This is the actual fix for the task since VELOv1 would not be tradeable at all

swansontec (approved):

CHANGELOG.md
swansontec: Ah, so say that in the release notes instead?

PR #294, opened 2023-10-06
Whitelist plugins for ChangeHero plugin

peachbits:
We can't rely on their blockchain names to match ours so we need to whitelist them manually

---
  - https://app.asana.com/0/0/1205417413219566

### DIFF
--- a/src/swap/changehero.ts
+++ b/src/swap/changehero.ts
@@ -21,14 +21,14 @@ import {
 
 import {
   checkInvalidCodes,
-  getCodes,
+  getCodesWithTranscription,
   getMaxSwappable,
   InvalidCurrencyCodes,
   makeSwapPluginQuote,
   SwapOrder
 } from '../swap-helpers'
 import { convertRequest, getAddress } from '../util/utils'
-import { EdgeSwapRequestPlugin } from './types'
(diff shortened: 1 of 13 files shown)


src/swap/changehero.ts
```diff
@@ -134,11 +161,26 @@ export function makeChangeHeroPlugin(
...
+    // Supported chains must be whitelisted
+    if (
+      MAINNET_CODE_TRANSCRIPTION[request.fromWallet.currencyInfo.pluginId] ==
+        null ||
+      MAINNET_CODE_TRANSCRIPTION[request.toWallet.currencyInfo.pluginId] == null
+    ) {
+      throw new SwapCurrencyError(
+        swapInfo,
+        request.fromCurrencyCode,
+        request.toCurrencyCode
```
swansontec, line 174: `throw new SwapCurrencyError(swapInfo, request)`. Passing currency codes is deprecated.

swansontec (changes requested):
Tiny change needed.

src/swap/changehero.ts
peachbits: I forgot about that. Better fix them all!

swansontec (approved):

PR #296, opened 2023-10-19
Make Thorchain give estimate quotes

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205733442381833

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 0.22.0 (2023-10-24)
+
+- changed: Thorchain quotes to estimate rate
+
 ## 0.21.11 (2023-10-19)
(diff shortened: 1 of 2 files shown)


peachbits (approved):
This seems fine but I don't understand the rationale of keeping the non-estimate code flows that can't be used. There's no justification in this PR or the associated asana task and the changelog entry implies it's permanent.

paullinator:
> This seems fine but I don't understand the rationale of keeping the non-estimate code flows that can't be used. There's no justification in this PR or the associated asana task and the changelog entry implies it's permanent.

The justification in keeping the fixed rate flows is for when we can export both quotes and to let the user choose.

PR #306, opened 2023-11-23
Fix thorchain min amount return error

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205953809312301

### DIFF
--- a/src/swap/defi/thorchain.ts
+++ b/src/swap/defi/thorchain.ts
@@ -1,4 +1,4 @@
-import { add, div, gt, lt, mul, round, sub } from 'biggystring'
+import { add, div, gt, mul, round, sub } from 'biggystring'
 import {
   asArray,
   asBoolean,
@@ -276,7 +276,6 @@ interface CalcSwapParams {
   affiliateFeeBasis: string
   streamingInterval: number
   streamingQuantity: number
-  dontCheckLimits?: boolean
 }
 
 interface CalcSwapResponse {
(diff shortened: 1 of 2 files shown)


src/swap/defi/thorchain.ts
```diff
@@ -237,15 +245,19 @@ const asQuoteSwap = asObject({
...
-  // recommended_min_amount_in: asString, // "1440032",
+  recommended_min_amount_in: asString, // "1440032",
   router: asOptional(asString), // "0xD37BbE5744D730a1d98d8DC97c42F0Ca46aD7146",
   // slippage_bps: asNumber, // 92,
   // streaming_slippage_bps: asNumber, // 5,
   streaming_swap_blocks: asNumber, // 170,
   total_swap_seconds: asOptional(asNumber) // 1020,
 })
 
+const asQuoteSwapFull = asEither(asQuoteError, asQuoteSwap)
```
samholmes, line 256: Why don't we use this in `getQuote`?

samholmes (approved):
Approved with curiosity

src/swap/defi/thorchain.ts
paullinator: Good point. Really it's just a type and I could have extended QuoteSwap instead of making more cleaners. I think i'll push a fixup

PR #309, opened 2023-12-19
Paul/upgrade core2.0

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1206061725717248

### DIFF
--- a/src/swap/changehero.ts
+++ b/src/swap/changehero.ts
@@ -296,7 +296,7 @@ export function makeChangeHeroPlugin(
     )
 
     const spendInfo: EdgeSpendInfo = {
-      currencyCode: fromCurrencyCode,
+      tokenId: request.fromTokenId,
       spendTargets: [
         {
           nativeAmount: amountExpectedFromNative,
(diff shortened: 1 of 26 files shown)


src/swap/changehero.ts
```diff
@@ -306,15 +306,27 @@ export function makeChangeHeroPlugin(
...
+        destAsset: {
+          pluginId: request.toWallet.currencyInfo.pluginId,
+          tokenId: request.toTokenId,
+          nativeAmount: amountExpectedToNative
+        },
+        sourceAsset: {
+          pluginId: request.fromWallet.currencyInfo.pluginId,
+          tokenId: request.fromTokenId,
+          nativeAmount: amountExpectedFromNative
+        },
```
swansontec, line 327: Not a request for change, but more of an "oops, oh well":

The swap request uses the terms "from" and "to", but the swap action uses the terms "dest" and "source". It might have been sweeter to do `fromAsset: EdgeAsset` and `toAsset: EdgeAsset` in the swap action types, since it's not only shorter but also matches the existing naming.

swansontec (approved):

paullinator:
Added commit to rename source/dest to to/fromAsset.

paullinator:
Do not merge until edge-core-js 2.0 is published and this PR is modified to upgrade core

PR #310, opened 2024-01-01
Paul/pulse change now

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1206049090065300

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- added: Support Pulsechain through ChangeNow
+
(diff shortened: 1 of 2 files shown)


src/swap/changehero.ts
samholmes, line 1: How are you able to get away with changing this field without changing the type in the core?

samholmes (approved):
Whoops. My comments were for the dependency PR. You can ignore them or considered them feedback for the dependency (if relevant).

Otherwise the pulsechain one-liner is approved.

PR #312, opened 2024-01-31
Split package

swansontec:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

We don't want to refactor everything all at once, but it looks like XRP is the one causing the crashes on iOS 13.

---
  - https://app.asana.com/0/0/1206470834673677

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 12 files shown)


paullinator (approved):
Nice. this ended up simpler than i thought it would be.

PR #320, opened 2024-03-30
Paul/est dex

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1206844698127216

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 /coverage/
 /dist/
 /lib/
+/edge/
 
 # Package managers:
 node_modules/
--- a/src/swap/defi/defiUtils.ts
+++ b/src/swap/defi/defiUtils.ts
@@ -96,7 +96,7 @@ export const getEvmApprovalData = async (params: {
       gasPrice: '20'
     }
   )
-  return approveTx.data
+  return approveTx.data != null ? approveTx.data.replace('0x', '') : undefined
 }
 
 export const getEvmTokenData = async (params: {
--- a/src/swap/defi/lifi.ts
+++ b/src/swap/defi/lifi.ts
@@ -45,6 +45,8 @@ const asInitOptions = asObject({
   integrator: asOptional(asString, 'edgeapp')
 })
 
+/** Allow up to 20% slippage even for variable rate quotes */
+const MAX_SLIPPAGE = '0.2'
 const LIFI_SERVERS_DEFAULT = ['https://li.quest']
 const EXPIRATION_MS = 1000 * 60
 const EXCHANGE_INFO_UPDATE_FREQ_MS = 60000
@@ -125,7 +127,7 @@ const asExchangeInfo = asObject({
 const asEstimate = asObject({
   fromAmount: asNumberString, // "400000",
   toAmount: asNumberString, // "237318132569913",
-  toAmountMin: asNumberString, // "225452225941418",
+  // toAmountMin: asNumberString, // "225452225941418",
   approvalAddress: asString, // "0x1231DEB6f5749EF6cE6943a275A1D3E7486F4EaE",
   executionDuration: asNumber // 1168,
   // feeCosts: asArray(asFeeCost)
@@ -279,6 +281,7 @@ export function makeLifiPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
       fromAddress,
       toAddress,
       integrator,
+      slippage: MAX_SLIPPAGE,
       fee: affiliateFee
     })
     // Get current pool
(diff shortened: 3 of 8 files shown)


src/swap/defi/defiUtils.ts
samholmes, line 1: I thought I got all the incorrect memo stuff done in my PR here: https://github.com/EdgeApp/edge-exchange-plugins/pulls?q=is%3Apr+is%3Aclosed I must have missed this one.

Hmm, should the other exchange plugins be using hex, or is text correct? I just realized that they are using text instead of hex, so I'm unsure if any of those needed changing. What does the plugin typically do with text memos other then encode them as bytes? What text encoding is used?

test/testpartner.ts
```diff
@@ -86,16 +96,52 @@ async function main(): Promise<void> {
...
+    }
+  }
   await context.changeLogSettings({
     defaultLogLevel: 'info',
     sources: {}
   })
 
-  const account = await context.loginWithKey('bob', loginKey)
+  // const lk = await account.getLoginKey()
+  // console.log(`Login key: ${lk}`)
```
samholmes, line 144: Cleanup?

test/testpartner.ts
```diff
@@ -109,6 +155,8 @@ async function main(): Promise<void> {
   const avaxWallet = await account.waitForCurrencyWallet(avaxInfo?.id ?? '')
   const arrrWallet = await account.waitForCurrencyWallet(arrrInfo?.id ?? '')
   const maticWallet = await account.waitForCurrencyWallet(maticInfo?.id ?? '')
+  await ethWallet.changePaused(false)
+  await avaxWallet.changePaused(false)
```
samholmes, line 159: Why are these wallets un-paused? Comment?

samholmes (changes requested):
In commit 3888e409df64b33bfa30baf009e43f3a589e6719 it looks like we change some constants and config, but I don't see where the API call change happened. Could you point that out in a review-bar?

Otherwise, the rest of my feedback are to add more comments to the code to resolve questions.

.gitignore
```diff
@@ -6,6 +6,7 @@
 /coverage/
 /dist/
 /lib/
+/edge/
```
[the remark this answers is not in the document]
paullinator: They are files created by the Node instance of the core when running the testpartner.ts. These are the cloned repos and encrypted keys.

test/testpartner.ts
paullinator: I left that in there since if you want a real account to login quickly, you'll want to get the loginKey and uncommenting those lines is how you get it

test/testpartner.ts
paullinator: Wallets are now paused by default so you have to unpause them to use them.

src/swap/defi/thorchainDa.ts
```diff
@@ -86,17 +86,18 @@ const asThorSwapRoute = asObject({
   providers: asArray(asString),
   calldata: asUnknown,
   expectedOutput: asString,
-  expectedOutputMaxSlippage: asString,
-  expectedOutputUSD: asString,
-  expectedOutputMaxSlippageUSD: asString,
+  // expectedOutputMaxSlippage: asString,
+  // expectedOutputUSD: asString,
+  // expectedOutputMaxSlippageUSD: asString,
```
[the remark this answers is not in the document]
paullinator: They are no longer used. `expectedOutputMaxSlippage` was the only one used and we don't care about the max slippage anymore with variable quotes since we just quote the user the expected amount.

Note that convention is to comment out unused cleaner params so we could re-enable them in the future if needed.

src/swap/defi/lifi.ts
```diff
@@ -125,7 +127,7 @@ const asExchangeInfo = asObject({
 const asEstimate = asObject({
   fromAmount: asNumberString, // "400000",
   toAmount: asNumberString, // "237318132569913",
-  toAmountMin: asNumberString, // "225452225941418",
+  // toAmountMin: asNumberString, // "225452225941418",
```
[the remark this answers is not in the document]
paullinator: unused cleaner params should be commented, not removed. Same as for `expectedOutputMaxSlippage` we don't show the user the 'worse case' quote anymore since it's a variable quote

test/testpartner.ts
samholmes: Add a comment similar to this then:

```js
// if you want a real account to login quickly, you'll want to get the loginKey and uncommenting those lines is how you get it
```

test/testpartner.ts
samholmes: Should this be an argument to the CLI?

PR #332, opened 2024-06-21
Return txid as orderId for DEX swaps when calling approve()

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1206648078202433

### DIFF
--- a/src/util/swapHelpers.ts
+++ b/src/util/swapHelpers.ts
@@ -99,7 +99,7 @@ export async function makeSwapPluginQuote(
   const toNativeAmount = action?.toAsset.nativeAmount
   const destinationAddress = action?.payoutAddress
   const isEstimate = action?.isEstimate ?? false
-  const quoteId = action?.orderId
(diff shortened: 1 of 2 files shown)


src/util/swapHelpers.ts
```diff
@@ -160,6 +160,10 @@ export async function makeSwapPluginQuote(
           savedAction.orderUri = `${savedAction.orderUri}${tx.txid}`
       }
 
+      if (quoteId == null && (swapInfo.isDex ?? false)) {
```
swansontec, line 163: Doing `(swapInfo.isDex === true)` also avoids the lint error, but makes more sense when you read it.

swansontec (approved):
One optional change.

src/util/swapHelpers.ts
peachbits: agreed, will fix

PR #334, opened 2024-06-25
Implement 0xgasless swap

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207395251755815

### DIFF
--- a/src/swap/defi/0x/0xGasless.ts
+++ b/src/swap/defi/0x/0xGasless.ts
@@ -1,8 +1,6 @@
-import { add } from 'biggystring'
 import {
   EdgeAssetAction,
   EdgeCorePluginFactory,
-  EdgeNetworkFee,
   EdgeSwapApproveOptions,
   EdgeSwapInfo,
   EdgeSwapQuote,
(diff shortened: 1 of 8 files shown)


src/swap/defi/0x/0xGasless.ts
```diff
@@ -175,10 +177,42 @@ export const make0xGaslessPlugin: EdgeCorePluginFactory = opts => {
             throw new Error(`Swap failed: ${apiSwapStatus.reason ?? 'unknown'}`)
           }
 
+          const assetAction: EdgeAssetAction = {
+            assetActionType: 'swap'
+          }
+          const orderId = apiSwapSubmition.tradeHash
+          const {
+            publicAddress: toWalletAddress
+          } = await swapRequest.toWallet.getReceiveAddress({
```
swansontec, line 186: This is really freaking me out. We never seem to request an address from the `toWallet` before this line, so how on earth are the funds arriving there? It *seems* like the gasless swaps are only within the same wallet, so do we need to add a `toWallet.id === fromWallet.id` check?

swansontec (approved):

swansontec (approved):

PR #344, opened 2024-10-02
Add 0xGasless swap fee

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208440926008971


src/swap/defi/0x/0xGasless.ts
```diff
@@ -27,6 +27,9 @@ const swapInfo: EdgeSwapInfo = {
   supportEmail: 'support@edge.app'
 }
 
+const EDGE_FEE_PERCENTAGE = 0.0075
```
paullinator, line 30: this should be have an override that can come from env.json on the GUI. See how we do Thorchain

paullinator (changes requested):

paullinator (approved):

PR #350, opened 2024-11-08
Remove partial flag from thorchain and maya protocol

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208678306130075


swansontec (approved):
Clarify with Paul?

Maybe also just delete the `isEstimate` flag altogether, since we don't support / don't want streaming swaps.

PR #354, opened 2024-12-02
Rango: Add Solana support

Ikari-Shinji-re:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none
 
note : 

In my tests, selecting a non-native token from the Solana blockchain and entering a value below my maximum balance resulted in an "insufficient balance" error. However, using the "max" button worked without any issues.
 

---
  - https://app.asana.com/0/0/1208425702564664


peachbits (approved):
Looks good! I also tested the code with EVM and Solana swaps and it's working as expected.

PR #358, opened 2024-12-24
Use data from quote response

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1209031065260182


paullinator (approved):

src/swap/defi/thorchain/swapkit.ts
```diff
@@ -92,6 +89,9 @@ const asThorSwapRoute = asObject({
   // expectedOutputMaxSlippage: asString,
   // expectedOutputUSD: asString,
   // expectedOutputMaxSlippageUSD: asString,
+  transaction: asObject({
```
paullinator, line 92: Are you for sure sure that this isn't optional. There may not be a data payload for some types of swaps? Just make sure to test that this is true when sending EVM, UTXO, and Solana (if supported)

paullinator (approved):

src/swap/defi/thorchain/swapkit.ts
peachbits: Good catch! It's not there for UTXO. I added a fixup commit

paullinator (approved):

PR #360, opened 2025-01-02
Add Unizen DEX

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208752773501973

### DIFF
--- a/src/swap/defi/unizen.ts
+++ b/src/swap/defi/unizen.ts
@@ -1,0 +1,434 @@
+import contracts from '@unizen-io/unizen-contract-addresses/production.json'
+import { div } from 'biggystring'
+import {
+  asArray,
+  asMaybe,
+  asNumber,
+  asObject,
+  asString,
+  asTuple,
+  asValue
+} from 'cleaners'
+import {
+  EdgeCorePluginOptions,
+  EdgeLog,
+  EdgeMemo,
+  EdgeSpendInfo,
+  EdgeSwapInfo,
+  EdgeSwapPlugin,
+  EdgeSwapQuote,
+  EdgeSwapRequest,
+  EdgeTransaction,
+  JsonObject,
+  SwapCurrencyError
+} from 'edge-core-js/types'
+
+import {
+  getMaxSwappable,
+  makeSwapPluginQuote,
+  SwapOrder
+} from '../../util/swapHelpers'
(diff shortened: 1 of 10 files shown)


src/swap/defi/unizen.ts
```diff
@@ -0,0 +1,435 @@
...
+        .map(p => p.name)
+        .join(', ')}`,
+      minReceiveAmount: evmSingleChain[0].transactionData.info.amountOutMin
+    }
+  }
+
+  const crossChain = asMaybe(asCrossChainSwapQuote)(rawQuote)
+  if (crossChain != null) {
+    const memos: EdgeMemo[] = []
+    const memoValue = crossChain[0].transactionData.memo
```
swansontec, line 364: The `asCrossChainSwapQuote` is defined using `asArray`, but here we only access `crossChain[0]`. What if there are multiple tx's in the quote? Would the customer lose money, because the funds don't get fully swapped? I would feel safer either using `asTuple` to enforce a single item in the array, or somehow looping over the tx's to handle the weird case.

swansontec (approved):
Approved, but there are some questions and worries.

src/swap/defi/unizen.ts
```diff
@@ -0,0 +1,435 @@
...
+  // UTXO
+  bitcoin: { unizenId: -3980891822, unizenName: '', makeFee: makeUtxoFees },
+  bitcoincash: { unizenId: -174457306, unizenName: '', makeFee: makeUtxoFees },
+  litecoin: { unizenId: -33463083, unizenName: '', makeFee: makeUtxoFees },
+  dogecoin: { unizenId: -3143381382, unizenName: '', makeFee: makeUtxoFees }
+}
+
+const UNIZEN_CONTRACTS: {
+  [version: string]: StringMap
+} = JSON.parse(JSON.stringify(contracts))
```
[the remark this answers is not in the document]
peachbits: Oops! will fix before merging

src/swap/defi/unizen.ts
peachbits: It's an array of quotes and the best one is the first one. I'll change improve the cleaner name and change to asTuple to enforce it's existance

PR #363, opened 2025-01-09
Fix Rango amount metadata

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1209064047272954

### DIFF
--- a/src/swap/defi/rango.ts
+++ b/src/swap/defi/rango.ts
@@ -230,8 +230,6 @@ export function makeRangoPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
       throw new SwapCurrencyError(swapInfo, request)
     }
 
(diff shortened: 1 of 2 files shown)


src/swap/defi/rango.ts
```diff
@@ -501,7 +498,7 @@ export function makeRangoPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
           spendTargets: [
             {
               memo: evmTransaction.txData,
-              nativeAmount: ethNativeAmount,
+              nativeAmount: nativeAmount,
```
swansontec, line 501: This can just be `nativeAmount`, since the left & right are the same name.

swansontec (approved):

PR #370, opened 2025-02-21
Use depositWithExpiry for every EVM swap

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1209458441262827

### DIFF
--- a/src/swap/defi/thorchain/common.ts
+++ b/src/swap/defi/thorchain/common.ts
@@ -44,7 +44,7 @@ import {
   QueryParams
 } from '../../../util/utils'
 import { EdgeSwapRequestPlugin, MakeTxParams } from '../../types'
-import { getEvmApprovalData, getEvmTokenData } from '../defiUtils'
+import { getDepositWithExpiryData,getEvmApprovalData } from '../defiUtils'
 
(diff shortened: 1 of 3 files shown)


src/swap/defi/thorchain/common.ts
```diff
@@ -570,39 +570,42 @@ export function makeThorchainBasedPlugin(
...
           contractAddress: router,
           assetAddress: sourceTokenContractAddress,
           nativeAmount: fromNativeAmount
         })
-      } else {
-        memo = Buffer.from(memo).toString('hex')
       }
+
+      // Need to use ethers.js to craft a proper tx that calls Thorchain contract, then extract the data payload
+      memo = await getEvmTokenData({
```
Jon-edge, line 602: Let's rename this function now that it's more generic, the name is misleading now because it's not specific to tokens

Jon-edge (changes requested):
This seems to be along the lines of what I found during my research on the issue.

Small change requested to avoid future confusion.

Jon-edge (approved):

PR #377, opened 2025-04-15
Upgrade to 0x gasless v2

Jon-edge:
https://0x.org/docs/upgrading/upgrading_to_gasless_v2

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208400988810616

### DIFF
--- a/src/swap/defi/0x/0xGasless.ts
+++ b/src/swap/defi/0x/0xGasless.ts
@@ -36,6 +36,10 @@ export const make0xGaslessPlugin: EdgeCorePluginFactory = opts => {
   return {
     swapInfo,
     fetchSwapQuote: async (swapRequest): Promise<EdgeSwapQuote> => {
+      if (swapRequest.quoteFor === 'to') {
+        throw new Error('To quotes not supported in Gasless v2 API')
+      }
+
       // The fromWallet and toWallet must be of the same because the swap
       // service only supports swaps of the same network and for the same
       // account/address.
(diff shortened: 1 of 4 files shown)


src/swap/defi/0x/0xGasless.ts
```diff
@@ -71,36 +71,34 @@ export const make0xGaslessPlugin: EdgeCorePluginFactory = opts => {
...
-      // The plugin only supports gasless swaps, so if approval is required
-      // it must be gasless.
+      // Check if approval is required but gasless approval is not available
       if (
-        // eslint-disable-next-line @typescript-eslint/prefer-optional-chain
-        apiSwapQuote.approval != null &&
-        apiSwapQuote.approval.isRequired &&
-        !apiSwapQuote.approval.isGaslessAvailable
+        apiSwapQuote.issues?.allowance != null &&
+        apiSwapQuote.approval == null
```
samholmes, line 103: Are you sure this is right? I'm reading the migration guide and it looks like the logic is different. Could you double-check this? Maybe some comments explaining what each means specifically and a "why" of the if.

samholmes (changes requested):

src/swap/defi/0x/0xGasless.ts
```diff
@@ -71,36 +71,36 @@ export const make0xGaslessPlugin: EdgeCorePluginFactory = opts => {
         tokenId: swapRequest.fromTokenId
       })
 
-      // Amount request parameter/field name to use in the quote request
-      const amountField =
-        swapRequest.quoteFor === 'to' ? 'buyAmount' : 'sellAmount'
```
samholmes, line 76: We should handle the unsupported quote direction?

samholmes (approved):

PR #385, opened 2025-05-27
Update 0xGasless orderUri

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1209918996092092

### DIFF
--- a/src/swap/defi/0x/0xGasless.ts
+++ b/src/swap/defi/0x/0xGasless.ts
@@ -181,6 +181,12 @@ export const make0xGaslessPlugin: EdgeCorePluginFactory = opts => {
 
           const orderId: string = apiSwapStatus.transactions[0].hash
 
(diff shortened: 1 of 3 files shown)


src/swap/defi/0x/0xGasless.ts
```diff
@@ -181,6 +181,12 @@ export const make0xGaslessPlugin: EdgeCorePluginFactory = opts => {
 
           const orderId: string = apiSwapStatus.transactions[0].hash
 
+          let orderUri: string | undefined
+          const { transactionExplorer } = swapRequest.fromWallet.currencyInfo
+          if (transactionExplorer.includes('%s')) {
```
swansontec, line 186: The `if` isn't really needed, because the `.replace` is a no-op if the string isn't there.

PR #401, opened 2025-08-27
feat(lifi): add SUI blockchain support

samholmes:
- Add SUI to getParentTokenContractAddress() with native token address
- Add SUI to MAINNET_CODE_TRANSCRIPTION mapping
- Add SUI transaction request cleaner (asTransactionRequestSui)
- Implement SUI-specific transaction building logic
- Handle SUI transactions similar to Solana (no approval needed)
- Update CHANGELOG.md with feature addition

SUI is now supported for DEX swaps through the LI.FI plugin with chain ID 9270000000000000

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1210756559961635

### DIFF
--- a/src/swap/defi/lifi.ts
+++ b/src/swap/defi/lifi.ts
@@ -18,6 +18,7 @@ import {
   SwapBelowLimitError,
   SwapCurrencyError
 } from 'edge-core-js/types'
+import { base64 } from 'rfc4648'
 
 import { div18 } from '../../util/biggystringplus'
 import {
@@ -36,7 +37,12 @@ import {
   makeQueryParams,
   promiseWithTimeout
 } from '../../util/utils'
-import { asNumberString, EdgeSwapRequestPlugin, StringMap } from '../types'
+import {
+  asNumberString,
+  EdgeSwapRequestPlugin,
+  MakeTxParams,
+  StringMap
+} from '../types'
 import { getEvmApprovalData, WEI_MULTIPLIER } from './defiUtils'
 
 const pluginId = 'lifi'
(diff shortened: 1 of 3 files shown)


src/swap/defi/lifi.ts
```diff
@@ -405,6 +415,43 @@ export function makeLifiPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
 
         break
       }
+      case 'sui': {
+        // SUI uses pre-built transactions
+        const { data } = asTransactionRequestSui(transactionRequestRaw)
+
+        spendInfo = {
```
peachbits, line 422: Create a makeTxParams instead of spendInfo so we can utilize otherMethods.makeTx to handle the unsigned data blob

peachbits (changes requested):

samholmes:
## Updated Implementation

I've addressed the feedback by implementing the makeTxParams approach as requested. Here's what changed:

### Implementation Details

1. **Using makeTxParams** - The SUI case now returns a SwapOrder with makeTxParams instead of spendInfo
2. **Embedded unsignedTx** - The pre-built transaction from LI.FI is converted to Uint8Array and embedded in makeTxParams
3. **Custom handling required** - Added detailed comments explaining that the SUI wallet needs to:
   - Extract the unsignedTx from makeTxParams
   - Call otherMethods.makeTx(unsignedTx, metadata) directly
   - This is necessary because the standard plugin infrastructure passes makeTxParams as a single argument, but SUI's makeTx expects (unsignedTx, metadata)

### Note on Architecture Mismatch

There's a fundamental mismatch between:
- **Plugin infrastructure**: Calls `otherMethods.makeTx(makeTxParams)`
- **SUI implementation**: Expects `otherMethods.makeTx(unsignedTx: Uint8Array, metadata?: MakeTxMetadata)`

The current implementation embeds the unsignedTx in makeTxParams with a type cast to `any` to allow this custom field. The SUI wallet will need to handle this specially until the plugin infrastructure can be updated to support this use case.

### Testing

Once the SUI wallet implements the custom makeTxParams handling, swaps should work correctly through LI.FI's Aftermath and Bluefin7k DEX integrations.

peachbits (approved):

PR #403, opened 2025-09-12
Fix Bridgeless Zano deposits

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

https://github.com/EdgeApp/edge-currency-accountbased/pull/978
https://github.com/EdgeApp/react-native-zano/pull/8

### Description

 none

---
  - https://app.asana.com/0/0/1211349016941708

### DIFF
--- a/src/swap/defi/bridgeless.ts
+++ b/src/swap/defi/bridgeless.ts
@@ -15,10 +15,12 @@ import {
   EdgeCurrencyWallet,
   EdgeFetchFunction,
   EdgeSpendInfo,
+  EdgeSwapApproveOptions,
   EdgeSwapInfo,
   EdgeSwapPlugin,
   EdgeSwapQuote,
   EdgeSwapRequest,
+  EdgeSwapResult,
   EdgeToken,
   EdgeTokenId,
   EdgeTxActionSwap,
@@ -45,6 +47,7 @@ const swapInfo: EdgeSwapInfo = {
 
 const BASE_URL = 'https://rpc-api.node0.mainnet.bridgeless.com'
 const ORDER_URL = 'https://tss1.mainnet.bridgeless.com'
+const AUTO_BOT_URL = 'https://autobot-wusa1.edge.app'
 
 const EDGE_PLUGINID_CHAINID_MAP: Record<string, string> = {
   bitcoin: '0',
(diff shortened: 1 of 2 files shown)


src/swap/defi/bridgeless.ts
```diff
@@ -353,7 +356,50 @@ export function makeBridgelessPlugin(
...
+          ) {
+            swapResult.transaction.savedAction.orderId.replace('{{TXID}}', txid)
+
+            const [
+              chainId,
+              ,
+              txNonce
+            ] = swapResult.transaction.savedAction.orderId.split('/')
+
+            opts.io
```
paullinator, line 381: await this. if it fails throw an error as the user should know that the swap will likely not complete. An error like

'Swap transaction sent but error creating withdrawal transaction. Contact [support email] if you do not receive funds within 2 hours'

src/swap/defi/bridgeless.ts
```diff
@@ -277,6 +277,7 @@ export function makeBridgelessPlugin(
 
         const spendInfo: EdgeSpendInfo = {
           otherParams: {
+            outputSort: 'targets',
```
paullinator, line 280: yikes. did we lose funds prior to this?

paullinator (changes requested):

src/swap/defi/bridgeless.ts
```diff
@@ -353,7 +356,50 @@ export function makeBridgelessPlugin(
...
+      const swapPluginQuote = await makeSwapPluginQuote(swapOrder)
+
+      const out = {
+        ...swapPluginQuote,
+        async approve(_opts?: EdgeSwapApproveOptions): Promise<EdgeSwapResult> {
+          const swapResult = await swapPluginQuote.approve(_opts)
+          const { txid } = swapResult.transaction
+
+          if (
+            swapResult.transaction.savedAction?.actionType === 'swap' &&
```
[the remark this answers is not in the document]
peachbits: It will always be true for this plugin but typescript demands we be sure because other savedAction types don't have orderId

src/swap/defi/bridgeless.ts
peachbits: I disagree. This will leave the user on the confirmation scene where they could slide to confirm again. It would be possible to burn the Zano assets multiple times.

src/swap/defi/bridgeless.ts
```diff
@@ -350,7 +357,47 @@ export function makeBridgelessPlugin(
...
+            ] = swapResult.transaction.savedAction.orderId.split('/')
+
+            await opts.io.fetch(`${AUTO_BOT_URL}/api/bridgeless`, {
+              method: 'PUT',
+              headers: {
+                'Content-Type': 'application/json'
+              },
+              body: JSON.stringify({ chainId, txHash: txid, txNonce })
+            })
+          }
```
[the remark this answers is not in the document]
paullinator: bugbot has a point. need to check for response.ok === true.

paullinator (changes requested):

paullinator (approved):

PR #404, opened 2025-09-17
Make CurrencyPluginIdSwapChainCodeMap keys optional

paullinator:
Update list of pluginId with missing values

### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1211387560996453


peachbits (changes requested):
The purpose of making them required to begin with is to force a developer adding a new currency plugin to consider existing swaps plugins and update them as well.

PR #433, opened 2026-02-04
update api

ex-peter-parker:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---

> [!NOTE]
> **Medium Risk**
> Changes request/response shapes and limit validation for a live swap provider, so mismatches with Exolix API expectations could break quoting or cause incorrect limit errors, especially for EVM/token swaps.
> 
> **Overview**
> Updates the Exolix swap integration to use Exolix’s newer network/address-based API parameters: requests now send `networkFrom`/`networkTo` (including an `evmGeneric` path with `evmChainId`) plus optional token `coinAddressFrom`/`coinAddressTo`, and the `/rate` response now requires a `rateId` that is passed through to `/transactions`.
> 
> Replaces the previous currency-code transcription path and the hardcoded $70k USD cap with min/max enforcement directly from Exolix (`minAmount`/`maxAmount` and `withdrawMin`/`withdrawMax`), throwing `SwapBelowLimitError`/`SwapAboveLimitError` accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f4ddc64a93a4df6158b4a279d0dfe161c9ef94c4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---
  - https://app.asana.com/0/0/1213596310518725

### DIFF
--- a/src/swap/central/exolix.ts
+++ b/src/swap/central/exolix.ts
@@ -1,4 +1,4 @@
-import { gt, lt, mul } from 'biggystring'
+import { gt, lt } from 'biggystring'
 import {
   asEither,
   asMaybe,
@@ -10,6 +10,7 @@ import {
 } from 'cleaners'
 import {
   EdgeCorePluginOptions,
+  EdgeCurrencyWallet,
   EdgeFetchResponse,
   EdgeMemo,
   EdgeSpendInfo,
@@ -23,12 +24,12 @@ import {
 } from 'edge-core-js/types'
 
 import { exolix as exolixMapping } from '../../mappings/exolix'
-import { div18 } from '../../util/biggystringplus'
+import { EdgeCurrencyPluginId } from '../../util/edgeCurrencyPluginIds'
 import {
   checkInvalidTokenIds,
   checkWhitelistedMainnetCodes,
   CurrencyPluginIdSwapChainCodeMap,
-  getCodesWithTranscription,
+  getContractAddresses,
   getMaxSwappable,
   InvalidTokenIds,
   makeSwapPluginQuote,
@@ -38,21 +39,15 @@ import {
 import {
   convertRequest,
   denominationToNative,
-  fetchRates,
   getAddress,
   memoType,
   nativeToDenomination
 } from '../../util/utils'
-import {
-  asRatesResponse,
-  EdgeSwapRequestPlugin,
-  RatesRespose,
-  StringMap
-} from '../types'
+import { EdgeSwapRequestPlugin, StringMap } from '../types'
 
 const pluginId = 'exolix'
 
-const swapInfo: EdgeSwapInfo = {
+export const swapInfo: EdgeSwapInfo = {
   pluginId,
   isDex: false,
   displayName: 'Exolix',
@@ -63,7 +58,6 @@ const asInitOptions = asObject({
   apiKey: asString
 })
 
-const MAX_USD_VALUE = '70000'
 const INVALID_TOKEN_IDS: InvalidTokenIds = {
   from: {
     polygon: [
@@ -78,6 +72,31 @@ const INVALID_TOKEN_IDS: InvalidTokenIds = {
     ]
   }
 }
+
+interface ExolixCommonQuoteParams {
+  networkFrom: string
+  networkTo: string
+  coinAddressFrom?: string
+  coinAddressTo?: string
+  networkFromChainId?: number
+  networkToChainId?: number
+  withdrawalAddress: string
+  withdrawalExtraId: string
+  refundAddress: string
+  refundExtraId: string
+  rateType: 'fixed' | 'float'
+  rateId?: string
+}
+
+type ExolixFromQuoteParams = ExolixCommonQuoteParams & {
+  amount: string
+}
+
+type ExolixToQuoteParams = ExolixCommonQuoteParams & {
+  withdrawalAmount: string
+}
+
+type ExolixQuoteParams = ExolixFromQuoteParams | ExolixToQuoteParams
 
 const addressTypeMap: StringMap = {
   digibyte: 'publicAddress',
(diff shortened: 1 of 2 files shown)


src/swap/central/exolix.ts
```diff
@@ -194,47 +239,48 @@ export function makeExolixPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
...
+      networkFromChainId,
+      coinAddressTo,
+      networkTo,
+      networkToChainId,
+      rateType: 'fixed',
+      withdrawalAddress: toAddress,
+      refundAddress: fromAddress,
+      refundExtraId: '',
+      withdrawalExtraId: '',
+      ...amount
```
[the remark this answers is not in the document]
samholmes: The concern seems valid here. Perhaps the fix is to use conditional spreads:

```ts
const quoteParams: ExolixQuoteParams = {
  ...(coinAddressFrom != null ? { coinAddressFrom } : {}),
  ...(networkFromChainId != null ? { networkFromChainId } : {})
```

src/swap/central/exolix.ts
```diff
@@ -99,21 +118,35 @@ while true; do
...
 const orderUri = 'https://exolix.com/transaction/'
 const uri = 'https://exolix.com/api/v2/'
 
 const expirationMs = 1000 * 60
 
 const asRateResponse = asObject({
   minAmount: asNumber,
+  maxAmount: asNumber,
   withdrawMin: asOptional(asNumber, 0),
+  withdrawMax: asOptional(asNumber, 0),
```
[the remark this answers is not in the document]
samholmes: Solution for this is to keep `withdrawMax` optional with no default and case the check under the condition where `withdrawMax` is present.

src/swap/central/exolix.ts
```diff
@@ -367,71 +422,6 @@ export function makeExolixPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
...
-        rates = asRatesResponse(reply)
-      } catch (e) {
-        log.error('Error fetching rates', String(e))
-        throw new Error('Error fetching rates')
-      }
-
-      const { exchangeRate } = rates.data[0]
-      if (exchangeRate == null) throw new SwapCurrencyError(swapInfo, request)
-
-      const usdValue = mul(exchangeAmount, exchangeRate)
```
samholmes: **Warning:** This PR materially changes Exolix quote behavior (new request shape, max/min handling, and 422 parsing) without corresponding automated tests.\n\n**Recommendation:** Add targeted tests for `quoteFor: from/to`, missing `withdrawMax`, and 422 minimum responses to prevent regressions.

src/swap/central/exolix.ts
```diff
@@ -99,21 +118,35 @@ while true; do
...
+}
+
 const orderUri = 'https://exolix.com/transaction/'
 const uri = 'https://exolix.com/api/v2/'
 
 const expirationMs = 1000 * 60
 
 const asRateResponse = asObject({
   minAmount: asNumber,
+  maxAmount: asNumber,
```
samholmes, line 143: **Issue:** The 422 fallback uses `asMaybe(asRateResponse)`, but this cleaner now requires fields like `maxAmount` and `rateId`. If Exolix returns a partial 422/minimum payload, parsing fails and this path throws `SwapCurrencyError` instead of returning a precise limit error.

**Recommendation:** Use a dedicated looser cleaner for 422 limit payloads, and reserve `asRateResponse` for successful quote payloads.

src/swap/central/exolix.ts
```diff
@@ -254,26 +310,25 @@ export function makeExolixPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
...
-      networkTo: quoteParams.coinToNetwork,
-      amount: quoteAmount,
-      withdrawalAddress: toAddress,
-      withdrawalExtraId: '',
-      refundAddress: fromAddress,
-      refundExtraId: '',
-      rateType: 'fixed'
+    const exchangeParams: ExolixQuoteParams = {
+      ...quoteParams,
+      rateId: rateResponse.rateId
```
samholmes, line 332: **Warning:** This change removes the prior local `$70k` USD guardrail. If that cap was a product/risk requirement, behavior has changed and large swaps now rely solely on partner-provided limits.

**Recommendation:** Confirm policy intent and keep a local cap if still required.

src/swap/central/exolix.ts
```diff
@@ -367,71 +422,6 @@ export function makeExolixPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
       const fixedOrder = await getFixedQuote(newRequest, userSettings)
       const fixedResult = await makeSwapPluginQuote(fixedOrder)
```
samholmes, line 427: **Warning:** This PR materially changes Exolix quote behavior (new request shape, max/min handling, and 422 parsing) without corresponding automated tests.

**Recommendation:** Add targeted tests for `quoteFor: from/to`, missing `withdrawMax`, and 422 minimum responses to prevent regressions.

samholmes, line 427: From the API requirements angle, this PR does move the integration in the right direction in a few meaningful ways: it updates the request shape to use chain-aware identifiers (`evmGeneric` plus chain IDs for EVM assets), includes token contract addresses, and adds explicit destination-side quoting support via `quoteFor: 'to'`. Those are all improvements toward the chain/token identification and bi-directional quoting requirements.

The highest-priority remaining concern is whether the new destination-side quoting path is fully reliable in practice. Because `quoteFor: 'to'` is new here and limit handling now depends on `withdrawMin` / `withdrawMax`, I would want confidence that Exolix always returns the necessary limit fields or that the plugin tolerates missing fields without manufacturing false over-limit failures.

Beyond that, the other material API-requirements items are provider capabilities that matter to user experience and ops, but are not really proven by this plugin diff: an unauthenticated status page, a transaction-status API by `orderId`, and reporting data that can be matched back to Edge orders. I would not block this PR on every last detail of error-shape perfection or extra hardening if the Exolix team has already confirmed those behaviors, but before treating the integration as compliant with the API requirements, I would want those core capabilities explicitly confirmed somewhere in the PR or supporting notes.

samholmes (changes requested):

src/swap/central/exolix.ts
ex-peter-parker: I confirm this — it was necessary for the old functionality; now all limits are calculated automatically.

src/swap/central/exolix.ts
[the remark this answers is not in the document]
ex-peter-parker: fixed

src/swap/central/exolix.ts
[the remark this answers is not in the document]
ex-peter-parker: fixed

PR #439, opened 2026-02-18
feat: add sonic to rango

RyukTheCoder:
Added Sonic to Rango and tested the following routes:
Sonic.S -> Sonic.wS

### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none


j0ntz:
Consolidated into #456, which combines the Sonic (#439) and Sui (#438) Rango changes. Note: the Sui changes from this set already landed in `master` (shipped in `2.46.0`/`2.44.0` as "(Rango) SUI support"), so #456's only net change is the Sonic mapping. Please track #456 going forward.

PR #449, opened 2026-03-27
Feat: Changelly integration

changelly:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

PR's contains no dependencies.

### Description

PR introduces new Changelly Relay API matching the actual JSON-RPC envelope.

---

> [!NOTE]
> **Medium Risk**
> Adds a new centralized swap integration that generates spend transactions and makes authenticated JSON-RPC calls to Changelly; correctness depends on API behavior, mapping accuracy, and error translation (limits/permission).
> 
> **Overview**
> Adds **Changelly** as a new centralized swap plugin (`src/swap/central/changelly.ts`) that fetches supported assets, requests fixed-rate quotes, creates swap transactions, and translates API errors into `SwapBelowLimitError`/`SwapAboveLimitError`/`SwapPermissionError`.
> 
> Introduces Changelly currency/network transcription mappings (`scripts/mappings/changellyMappings.ts` → auto-generated `src/mappings/changelly.ts`) and registers the plugin in `src/index.ts`. Updates `CHANGELOG.md` with an unreleased entry for Changelly support.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d3d058d75f40981109ac9ab6dd845b30fcfac1f3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/swap/central/changelly.ts
```diff
@@ -0,0 +1,631 @@
...
+        addressTypeMap[request.toWallet.currencyInfo.pluginId]
+      )
+    ])
+
+    const {
+      fromCurrencyCode,
+      toCurrencyCode
+    } = getCodesWithTranscription(request, MAINNET_CODE_TRANSCRIPTION)
+
+    const fromTicker =
```
peachbits: getCodesWithTranscription() is the wrong abstraction for this integration. That helper is useful for providers that model assets as separate chainCode + tokenCode values and mostly key off currency codes. Changelly does not work that way here. It exposes a single provider asset id per supported asset, and this PR only passes the mainnet transcription map, so the returned Changelly IDs would land in fromMainnetCode / toMainnetCode anyway. The quote path never uses those values, so the mapped Changelly IDs are effectively ignored.

There is a second problem in the same area: falling back to fromCurrencyCode.toLowerCase() / toCurrencyCode.toLowerCase() when the asset is not found in the provider cache is unsafe. If the cache lookup misses, the plugin can silently request a quote for a different asset than the one the user selected. ETH is the obvious example, since many distinct chains use that currency code.

A better approach would be to build the cache as a mapping from Changelly-supported assets to local pluginId / tokenId pairs using the provider payload, especially contractAddress. Convert the Changelly supported assets response to a `ChainCodeTickerMap` and then use that with the `getChainAndTokenCodes` utility. Using the returned `fromCurrencyCode` and `toCurrencyCode` from that function would result in the correct Changelly IDs for the quote. Changehero, Changenow, Letsexchange, and Sideshift have all adopted this pattern.

peachbits (changes requested):

src/swap/central/changelly.ts
```diff
@@ -0,0 +1,631 @@
...
+      .sort((a, b) => {
+        return (
+          new Date(b.lastLogin).getTime() - new Date(a.lastLogin).getTime()
+        )
+      })
+
+    if (profiles.length === 0) throw new Error('Unable to detect user params')
+
+    return {
+      userId: profiles[0].userId,
```
peachbits: Let's use initOptions (passed in via EdgeCorePluginOptions) instead of using disklet. We can pass in any arbitrary fields necessary.

peachbits (changes requested):

src/swap/central/changelly.ts
changelly: Requested changes were provided with https://github.com/EdgeApp/edge-exchange-plugins/pull/449/changes/b3c7bd5631e4fe1f34f64569221ecbb92ca10b60

src/swap/central/changelly.ts
changelly: These options are integral to our new 'relay' strategy. We have opted to bypass endpoint authentication via provided credentials in favor of authenticating edge users through non-sensitive data stored locally(zero-config authentication).

peachbits: The data accessed in this blob are unique to the user and don't identify them as Edge users. Also, it passes the username of any account that attempts a swap to Changelly, which is non-starter. We don't associate usernames with swaps and won't pass them along. Instead, ignore the local disklet completely and accept the same string the Changelly reports-server PR is expecting "edge-app".

changelly: We've rolled out usage of a disklet from our client. Now it's pseudo-sequence that is related to a one, used in reports server.

src/swap/central/changelly.ts
```diff
@@ -0,0 +1,607 @@
...
+            : request.fromTokenId
+          const direction = reverseQuote ? 'to' : undefined
+
+          if (msg.includes('Minimal')) {
+            const minFrom = limits.min?.from
+            if (minFrom != null) {
+              const minNativeAmount = denominationToNative(
+                wallet,
+                minFrom,
+                tokenId
```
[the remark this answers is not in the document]
peachbits: should fix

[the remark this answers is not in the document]
peachbits: worth safe-guarding against if the API returns greater precision than what is actually possible

peachbits (changes requested):
I added a fixup commit with the preferred handling of api secrets. We can obfuscate them in the env.json like I did in this PR: https://github.com/EdgeApp/edge-react-gui/pull/6028 This adds obfuscation to the envconfig for use in CHANGELLY_INIT

src/swap/central/changelly.ts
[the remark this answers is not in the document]
changelly: Resolved in [86b41f4](https://github.com/EdgeApp/edge-exchange-plugins/pull/449/changes/86b41f4febaece299065fa433d6e5e2c9a9a660d)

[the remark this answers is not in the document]
changelly: Resolved in [86b41f4](https://github.com/EdgeApp/edge-exchange-plugins/pull/449/changes/86b41f4febaece299065fa433d6e5e2c9a9a660d)

peachbits:
Master has moved substantially since this branch was cut (npm conversion in 2.47.0, several releases since), so rather than asking for another rebase here I've re-cut the integration onto latest master in #471 as a fresh commit with co-authorship credit.

It carries this branch's head (86b41f4) plus a few small changes on top:

- the JSON-RPC id string now comes from a new `partnerId` initOption (Edge passes `edge-app`) rather than a hardcoded literal
- a guard that throws when `apiKey` or `partnerId` is empty/missing, so an unconfigured build cleanly runs without the plugin
- min/max limit amounts convert to native units before integer truncation (sub-unit limits like 0.0025 BTC previously collapsed to "0")
- named `makeChangellyPlugin` export + lint/import-order normalization

Keeping this PR open for reference — further review will happen on #471.

peachbits:
Replaced by https://github.com/EdgeApp/edge-exchange-plugins/pull/471

PR #457, opened 2026-06-08
Add NYM swap provider

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

Integrate NYM as a centralized swap provider (`pluginId: nym`).

Asana: https://app.asana.com/1/9976422036640/project/1213880789473005/task/1215306512581867

NYM provides its own swap liquidity, so every swap must have the NYM asset on one side. The plugin gates this locally before hitting the API to avoid offering NYM (and rate-limiting it) on unrelated pairs.

NYM exposes an Edge-shaped partner API, which keeps the plugin thin:

- **Native-unit amounts** — `sourceAmount`/`destinationAmount` are already in the chain's smallest unit (matching Edge's `nativeAmount`), so no denomination conversion is needed in either direction.
- **Edge-style asset references** — `{ chainNetwork, chainId?, tokenId? }`. `chainNetwork` is the network-family name (mapped from the Edge `pluginId`), `chainId` comes from the wallet's `evmChainId`, and `tokenId` is the **0x contract address** (not the Edge tokenId). Ethereum mainnet is represented as `chainNetwork: 'ethereum'` + `chainId: 1`.
- **Limits** — the quote response returns native-unit min/max for both sides; the plugin enforces them as `SwapBelowLimitError` / `SwapAboveLimitError`.
- **Errors** — NYM's JSON error bodies (`{ errors: [{ error, message }] }` and `{ error }`) are parsed into `SwapCurrencyError` / `SwapPermissionError`, with rate-limit/other failures surfaced as transient errors.

Files:
- `src/swap/central/nym.ts` — the provider.
- `src/mappings/nym.ts` — `EdgeCurrencyPluginId → chainNetwork`, scoped to the current livenet asset list (BTC, Ethereum ETH/USDT/USDC/ERC-20 NYM, native NYM on `nyx`).
- `src/index.ts` — registers `nym: makeNymPlugin`.

**Livenet.** Base URL points at the NYM production API (`https://nym-swap-api.nymtech.cc`) and the order-tracking link at `https://swap.nym.com/orderStatus/`. The mapping reflects the current `GET /currencies` list and should be expanded here as NYM enables more chains. The init option is `{ apiKey }` (header `x-api-key`).

**Verified end-to-end in the iOS simulator** (edge-react-gui, linked via updot): ran a real mainnet swap My Ether 4 (USDT on Ethereum) → My Nym (native NYM on Nyx). The quote resolved "Powered by NYM" (6.006 USDT → 326.99 NYM), slide-to-confirm broadcast the on-chain payin (max network fee 0.0002709 ETH), and the order entered processing ("Congratulations! Your exchange is being processed", tx category `Exchange: To NYM`, 6.006 USDT → 326.993333 NYM fixed quote). Also validated the raw livenet API directly (`GET /currencies`, `POST /quote` → `POST /order`).

**pluginId note:** the swap plugin is registered as `nymswap`, deliberately distinct from the `nym` *currency* plugin — edge-core-js keys all plugins in one namespace, so reusing `nym` silently dropped the swap plugin (mirrors how the Thorchain swap plugin `thorchain` differs from the `thorchainrune` currency). This was caught by the in-sim test.

---

> [!NOTE]
> **Medium Risk**
> New third-party swap integration that creates real pay-in orders and handles API keys and user addresses, though it follows existing centralized-plugin patterns and includes local pair gating and trusted order URLs.
> 
> **Overview**
> Adds a **NYM** centralized swap provider registered as `nymswap` (separate from the `nym` currency plugin), with a CHANGELOG entry.
> 
> The plugin talks to NYM’s livenet partner API (`quote` → `order`) using an `apiKey` init option. Swaps are only offered when **NYM is on one side** of the pair; supported chains are mapped in `src/mappings/nym.ts` (currently Bitcoin, Ethereum mainnet/tokens, and native NYM on Nyx). Quotes use native-unit amounts and enforce min/max limits; API errors map to `SwapCurrencyError` / `SwapPermissionError` where appropriate.
> 
> **Max-swap** uses a quote-only probe so `getMaxSwappable` does not create and abandon live NYM orders. Order status links are built from a fixed `ORDER_URI` rather than the partner’s `statusUrl`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9480ffb1cc4315f4126eb8074ec350eda14b4c56. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/swap/central/nym.ts
+++ b/src/swap/central/nym.ts
@@ -1,0 +1,432 @@
+import { gt, lt } from 'biggystring'
+import {
+  asArray,
+  asDate,
+  asMaybe,
+  asObject,
+  asOptional,
+  asString
+} from 'cleaners'
+import {
+  EdgeCorePluginOptions,
+  EdgeCurrencyWallet,
+  EdgeMemo,
+  EdgeSpendInfo,
+  EdgeSwapInfo,
+  EdgeSwapPlugin,
+  EdgeSwapQuote,
+  EdgeSwapRequest,
+  SwapAboveLimitError,
+  SwapBelowLimitError,
+  SwapCurrencyError,
+  SwapPermissionError
+} from 'edge-core-js/types'
+
+import { nym as nymMapping } from '../../mappings/nym'
+import { EdgeCurrencyPluginId } from '../../util/edgeCurrencyPluginIds'
+import {
+  checkInvalidTokenIds,
+  checkWhitelistedMainnetCodes,
+  CurrencyPluginIdSwapChainCodeMap,
+  ensureInFuture,
+  getContractAddresses,
+  getMaxSwappable,
+  makeSwapPluginQuote,
+  mapToRecord,
+  SwapOrder
+} from '../../util/swapHelpers'
+import { convertRequest, getAddress, memoType } from '../../util/utils'
+import { EdgeSwapRequestPlugin } from '../types'
+
+// Swap plugin id. Distinct from the `nym` *currency* plugin id (edge-core-js
+// keys all plugins in one namespace), mirroring how the Thorchain swap plugin
+// (`thorchain`) differs from the `thorchainrune` currency plugin.
+const pluginId = 'nymswap'
+
+// Edge pluginId of the NYM *asset/currency*. NYM provides its own liquidity, so
+// every swap must have the NYM asset on one side (enforced below).
+const NYM_PLUGIN_ID = 'nym'
+
+export const swapInfo: EdgeSwapInfo = {
+  pluginId,
+  isDex: false,
+  displayName: 'NYM',
+  supportEmail: 'support@nymtech.net'
+}
+
+const asInitOptions = asObject({
+  apiKey: asString
+})
+
+// NYM's production ("livenet") partner API. Request/response shapes and the
+// quote/order paths are identical to the earlier testnet backend; only the host
+// differs.
+const NYM_API_BASE = 'https://nym-swap-api.nymtech.cc'
+// User-facing order-tracking page. Built from this trusted constant rather than
+// the partner-supplied `statusUrl`, so a compromised upstream cannot inject an
+// attacker-controlled host/scheme into the saved swap action.
+const ORDER_URI = 'https://swap.nym.com/orderStatus/'
+
+const MAINNET_CODE_TRANSCRIPTION: CurrencyPluginIdSwapChainCodeMap = mapToRecord(
+  nymMapping
+)
+
+/**
+ * NYM asset reference, mirroring Edge's own asset model. `chainNetwork` is the
+ * NYM network-family name, `chainId` is the EVM chain id (omitted for non-EVM
+ * chains), and `tokenId` is the 0x contract address (null for a native asset).
+ */
+interface NymAssetRef {
+  chainNetwork: string
+  chainId?: number
+  tokenId: string | null
+}
+
+const asNymQuote = asObject({
+  quoteId: asString,
+  sourceAmount: asString,
+  destinationAmount: asString,
+  rate: asString,
+  expiresAt: asDate,
+  minSourceAmount: asString,
+  maxSourceAmount: asString,
+  minDestinationAmount: asString,
+  maxDestinationAmount: asString
+})
+
+const asNymOrder = asObject({
+  orderId: asString,
+  status: asString,
+  payinAddress: asString,
+  payinExtraId: asOptional(asString),
+  expiresAt: asDate
+  // `statusUrl` is intentionally not consumed: the order-tracking URL is built
+  // from the trusted ORDER_URI constant to avoid trusting a partner-supplied
+  // host/scheme.
+})
+
+// Error bodies come in two shapes:
+//   { errors: [{ error: 'InvalidRequest' | 'AssetNotSupported', message }] }
+//   { error: 'Quote rate limit exceeded ...' }
+const asNymErrorArray = asObject({
+  errors: asArray(asObject({ error: asString, message: asOptional(asString) }))
+})
+const asNymSimpleError = asObject({ error: asString })
+
+/**
+ * Builds a NYM asset reference from an Edge wallet. `contractAddress` is the
+ * token's 0x contract address (undefined for a native asset). Returns null if
+ * the wallet's chain is not mapped to a NYM `chainNetwork`.
+ */
+const getAssetRef = (
+  wallet: EdgeCurrencyWallet,
+  contractAddress: string | undefined
+): NymAssetRef | null => {
+  const chainNetwork =
+    MAINNET_CODE_TRANSCRIPTION[
+      wallet.currencyInfo.pluginId as EdgeCurrencyPluginId
+    ]
+  if (chainNetwork == null) return null
+
+  const evmChainId = wallet.currencyInfo.evmChainId
+  return {
+    chainNetwork,
+    ...(evmChainId != null ? { chainId: evmChainId } : {}),
+    tokenId: contractAddress ?? null
+  }
+}
+
+export function makeNymPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
+  const { io, log } = opts
+  const { apiKey } = asInitOptions(opts.initOptions)
+  const fetchCors = io.fetchCors ?? io.fetch
+
+  const headers = {
+    'Content-Type': 'application/json',
+    Accept: 'application/json',
+    'x-api-key': apiKey
+  }
+
+  /**
+   * Maps a non-OK NYM response to the appropriate swap error. Bodies are parsed
+   * when present; otherwise classification falls back to the status code.
+   */
+  const handleErrorResponse = async (
+    response: { status: number; text: () => Promise<string> },
+    request: EdgeSwapRequestPlugin,
+    stage: 'quote' | 'order'
+  ): Promise<never> => {
+    const text = await response.text().catch(() => '')
+    log.warn(`NYM ${stage} error ${response.status}: ${text}`)
+
+    let parsed: unknown
+    try {
+      parsed = JSON.parse(text)
+    } catch (e: unknown) {
+      parsed = undefined
+    }
+
+    const errorArray = asMaybe(asNymErrorArray)(parsed)
+    if (errorArray != null) {
+      const message = errorArray.errors
+        .map(e => e.message ?? e.error)
+        .join('; ')
+      // Region / geographic restriction.
+      if (/region|geo|country|blocked|jurisdiction/i.test(message)) {
+        throw new SwapPermissionError(swapInfo, 'geoRestriction')
+      }
+      // Unsupported asset or pair/direction (e.g. selling NYM to a UTXO chain).
+      throw new SwapCurrencyError(swapInfo, request)
+    }
+
+    // 403 with no parsable error body: treat as a permission restriction.
+    if (response.status === 403) {
+      throw new SwapPermissionError(swapInfo, 'geoRestriction')
+    }
+
+    // { error: 'message' } (e.g. 429 rate limiting) or anything else: transient.
+    const simpleError = asMaybe(asNymSimpleError)(parsed)
+    throw new Error(
+      `NYM ${stage} returned error code ${response.status}${
+        simpleError != null ? `: ${simpleError.error}` : ''
+      }`
+    )
+  }
+
+  /**
+   * Quote step: gate the pair, resolve asset refs + addresses, fetch a quote, and
+   * enforce the quote's native-unit min/max. Creates NO order, so it is safe to
+   * run as the `getMaxSwappable` probe, which would otherwise create (and abandon)
+   * a live NYM order on every max-swap request (see `fetchProbeOrder`).
+   *
+   * `enforceMax` is false only for the probe: a max-swap probe intentionally
+   * quotes the full pre-fee balance to discover the ceiling, so an above-limit
+   * balance must clamp via `getMaxSpendable` rather than throw `SwapAboveLimitError`
+   * on the raw balance.
+   */
+  const fetchQuote = async (
+    request: EdgeSwapRequestPlugin,
+    enforceMax: boolean
+  ): Promise<{
+    quote: ReturnType<typeof asNymQuote>
+    refundAddress: string
+    payoutAddress: string
+  }> => {
+    const { fromWallet, toWallet, quoteFor } = request
+
+    // NYM provides its own liquidity: one side of the swap must be NYM. Gate
+    // locally so unrelated pairs never reach (and rate-limit) the API.
+    if (
+      fromWallet.currencyInfo.pluginId !== NYM_PLUGIN_ID &&
+      toWallet.currencyInfo.pluginId !== NYM_PLUGIN_ID
+    ) {
+      throw new SwapCurrencyError(swapInfo, request)
+    }
+
+    // NYM identifies tokens by their 0x contract address, not the Edge tokenId.
+    const { fromContractAddress, toContractAddress } = getContractAddresses(
+      request
+    )
+
+    const from = getAssetRef(fromWallet, fromContractAddress)
+    const to = getAssetRef(toWallet, toContractAddress)
+    if (from == null || to == null) {
+      throw new SwapCurrencyError(swapInfo, request)
+    }
+
+    // Grab addresses:
+    const [refundAddress, payoutAddress] = await Promise.all([
+      getAddress(fromWallet),
+      getAddress(toWallet)
+    ])
+
+    // NYM amounts are in native units (matching Edge's nativeAmount), so no
+    // denomination conversion is required.
+    const quoteBody =
+      quoteFor === 'to'
+        ? { from, to, destinationAmount: request.nativeAmount }
+        : { from, to, sourceAmount: request.nativeAmount }
+
+    const quoteResponse = await fetchCors(
+      `${NYM_API_BASE}/api/partner/v1/quote`,
+      {
+        method: 'POST',
+        headers,
+        body: JSON.stringify(quoteBody)
+      }
+    )
+
+    if (!quoteResponse.ok) {
+      await handleErrorResponse(quoteResponse, request, 'quote')
+    }
+
+    const quote = asNymQuote(await quoteResponse.json())
+
+    // Enforce the min/max limits returned by the quote against the user's
+    // requested amount (NYM returns native-unit limits for both sides).
+    // Compare `request.nativeAmount`, not the quote's echoed amount, so a
+    // clamped quote that omits an error still trips the limit error.
+    const [min, max, side] =
+      quoteFor === 'to'
+        ? ([
+            quote.minDestinationAmount,
+            quote.maxDestinationAmount,
+            'to'
+          ] as const)
+        : ([quote.minSourceAmount, quote.maxSourceAmount, 'from'] as const)
+    if (lt(request.nativeAmount, min)) {
+      throw new SwapBelowLimitError(swapInfo, min, side)
+    }
+    if (enforceMax && gt(request.nativeAmount, max)) {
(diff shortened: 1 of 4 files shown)


src/swap/central/nym.ts
```diff
@@ -0,0 +1,361 @@
...
+    stage: 'quote' | 'order'
+  ): Promise<never> => {
+    const text = await response.text().catch(() => '')
+    log.warn(`NYM ${stage} error ${response.status}: ${text}`)
+
+    let parsed: unknown
+    try {
+      parsed = JSON.parse(text)
+    } catch (e) {
+      parsed = undefined
```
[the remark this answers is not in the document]
j0ntz: Fixed in 5cc72ff — changed to `catch (e: unknown)` per the project TypeScript catch-typing convention.

src/swap/central/nym.ts
```diff
@@ -0,0 +1,361 @@
...
+    const from = getAssetRef(fromWallet, fromContractAddress)
+    const to = getAssetRef(toWallet, toContractAddress)
+    if (from == null || to == null) {
+      throw new SwapCurrencyError(swapInfo, request)
+    }
+
+    // Grab addresses:
+    const [refundAddress, payoutAddress] = await Promise.all([
+      getAddress(fromWallet),
+      getAddress(toWallet)
```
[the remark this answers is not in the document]
j0ntz: Fixed in 5cc72ff — added an `addressTypeMap` ({ zcash: 'transparentAddress' }) and pass it to `getAddress` for both the refund and payout addresses, matching the other central plugins that support ZEC.

src/swap/central/nym.ts
```diff
@@ -0,0 +1,367 @@
...
+      memos,
+      networkFeeOption: 'high',
+      assetAction: {
+        assetActionType: 'swap'
+      },
+      savedAction: {
+        actionType: 'swap',
+        swapInfo,
+        orderId: order.orderId,
+        orderUri: order.statusUrl ?? ORDER_URI + order.orderId,
```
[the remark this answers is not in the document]
j0ntz: Good catch — fixed. The order-tracking URL is now built from a trusted constant (`ORDER_URI + order.orderId`) and the partner-supplied `statusUrl` is no longer consumed, so a compromised upstream cannot inject an attacker-controlled host/scheme into `savedAction.orderUri`. This matches how the other central plugins (e.g. sideshift) construct their own order URLs.

src/swap/central/nym.ts
```diff
@@ -0,0 +1,367 @@
...
+      const request = convertRequest(req)
+
+      checkWhitelistedMainnetCodes(
+        MAINNET_CODE_TRANSCRIPTION,
+        request,
+        swapInfo
+      )
+
+      const newRequest = await getMaxSwappable(fetchSwapQuoteInner, request)
+      const swapOrder = await fetchSwapQuoteInner(newRequest)
```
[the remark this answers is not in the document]
j0ntz: Acknowledged. This is the canonical central-swap pattern in this repo: `getMaxSwappable` probes `fetchSwapQuoteInner` to compute the spendable amount, and every order-creating central plugin (sideshift, exolix, letsexchange, changenow) takes the same shape — the probe's order is left unused and expires. Keeping NYM consistent with that reviewed convention rather than introducing a bespoke max-handling path (which would compute getMaxSpendable against a placeholder address and isn't covered by tests). The throwaway order simply expires at the partner; noting it as a known tradeoff. Open to a follow-up that defers order creation across all central plugins if the team prefers.

src/swap/central/nym.ts
```diff
@@ -0,0 +1,375 @@
...
+          'to'
+        )
+      }
+    } else {
+      if (lt(quote.sourceAmount, quote.minSourceAmount)) {
+        throw new SwapBelowLimitError(swapInfo, quote.minSourceAmount, 'from')
+      }
+      if (gt(quote.sourceAmount, quote.maxSourceAmount)) {
+        throw new SwapAboveLimitError(swapInfo, quote.maxSourceAmount, 'from')
+      }
```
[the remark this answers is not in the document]
j0ntz: Fixed — the min/max checks now compare the user's `request.nativeAmount` (the quoted side) against the limit fields instead of the quote's echoed amount, so a clamped quote that omits an error still trips SwapBelowLimitError / SwapAboveLimitError.

j0ntz:
## 📸 Test evidence — real NYM swap (Sepolia ETH → NYM) executed on iOS sim

**agent proof 1215306512581867 01 pair sepolia to nym**

agent proof 1215306512581867 01 pair sepolia to nym

**agent proof 1215306512581867 02 nym quote powered by**

agent proof 1215306512581867 02 nym quote powered by

**agent proof 1215306512581867 03 swap success**

agent proof 1215306512581867 03 swap success

**agent proof 1215306512581867 04 tx details to nym**

agent proof 1215306512581867 04 tx details to nym

j0ntz:
## 📸 Test evidence: livenet USDT to NYM swap on iOS sim

**livenet quote powered by nym**

livenet quote powered by nym

**swap success scene**

swap success scene

**tx details to nym**

tx details to nym

src/swap/central/nym.ts
```diff
@@ -0,0 +1,378 @@
...
+  ): Promise<SwapOrder> => {
+    const { fromWallet, toWallet, quoteFor } = request
+
+    // NYM provides its own liquidity: one side of the swap must be NYM. Gate
+    // locally so unrelated pairs never reach (and rate-limit) the API.
+    if (
+      fromWallet.currencyInfo.pluginId !== NYM_PLUGIN_ID &&
+      toWallet.currencyInfo.pluginId !== NYM_PLUGIN_ID
+    ) {
+      throw new SwapCurrencyError(swapInfo, request)
```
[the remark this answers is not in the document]
j0ntz: Verified against the live NYM API, this is not a functional gap. The API enforces "one side must be NYM" where NYM is the native asset on chainNetwork `nyx`; it rejects pairs where NYM appears only as the ERC-20 (0x525A8F6F3Ba4752868cde25164382BfbaE3990e1) on ethereum. Both `BTC -> NYM(erc20)` and `ETH -> NYM(erc20)` return `InvalidRequest: Unsupported swap pair: one side must be NYM`. So the local gate (native NYM, pluginId `nym`, on one side) matches the API exactly and only rejects pairs the API also rejects, avoiding a doomed round-trip. The ERC-20 NYM is still reachable as the counter-asset when native NYM is the source: `NYM(nyx) -> NYM(erc20 ethereum)` quotes fine and the gate allows it (native NYM on one side). No API-valid pair is lost, so no code change.

src/swap/central/nym.ts
```diff
@@ -0,0 +1,378 @@
...
+      const request = convertRequest(req)
+
+      checkWhitelistedMainnetCodes(
+        MAINNET_CODE_TRANSCRIPTION,
+        request,
+        swapInfo
+      )
+
+      const newRequest = await getMaxSwappable(fetchSwapQuoteInner, request)
+      const swapOrder = await fetchSwapQuoteInner(newRequest)
```
j0ntz, line 426: **CONFIRMED — correctness (severe).** `fetchSwapQuoteInner` both fetches a quote and places a real order (`POST /api/partner/v1/order`, lines 283-294) in one call. On a "swap max" request, `getMaxSwappable` (swapHelpers.ts:237-251) invokes this function as a probe with the raw untrimmed balance — that probe call already creates a real, live order on NYM's side. `fetchSwapQuoteInner` is then called again here with the trimmed amount, creating a **second** real order; only the second is returned to the user. The first order (with a real payin address) is silently abandoned server-side.

Every max-swap request against NYM burns two live orders and two full quote+order round trips instead of one. Consider splitting quote-fetching from order-creation into separate inner functions so `getMaxSwappable`'s probe only needs the quote step.

src/swap/central/nym.ts
```diff
@@ -0,0 +1,378 @@
...
+  }
+
+  const fetchSwapQuoteInner = async (
+    request: EdgeSwapRequestPlugin
+  ): Promise<SwapOrder> => {
+    const { fromWallet, toWallet, quoteFor } = request
+
+    // NYM provides its own liquidity: one side of the swap must be NYM. Gate
+    // locally so unrelated pairs never reach (and rate-limit) the API.
+    if (
```
j0ntz, line 219: **CONFIRMED — correctness.** This gate only requires NYM to be on "one side" of the swap, which is trivially true even when `fromWallet === toWallet` (a native-NYM self-swap). Unlike this plugin, all 7 sibling central plugins (exolix, godex, changenow, changehero, nexchange, swapuz, letsexchange) call `checkInvalidTokenIds`, which includes an `isSameAsset` check that rejects self-swaps client-side. Here, a self-swap request would pass this check and hit NYM's live `/quote` and `/order` endpoints for a no-op swap.

Consider calling `checkInvalidTokenIds` (or an equivalent same-asset check) alongside the NYM-required-on-one-side gate.

src/swap/central/nym.ts
```diff
@@ -0,0 +1,378 @@
...
+          swapInfo,
+          quote.maxDestinationAmount,
+          'to'
+        )
+      }
+    } else {
+      if (lt(request.nativeAmount, quote.minSourceAmount)) {
+        throw new SwapBelowLimitError(swapInfo, quote.minSourceAmount, 'from')
+      }
+      if (gt(request.nativeAmount, quote.maxSourceAmount)) {
```
j0ntz: **CONFIRMED — correctness.** This check runs on `request.nativeAmount`, which during `getMaxSwappable`'s probe call (swapHelpers.ts:248-251) is the wallet's **raw, pre-fee balance** — not the fee-trimmed spendable amount computed afterward via `getMaxSpendable`. If `quote.maxSourceAmount` is below the raw balance but above the true fee-adjusted spendable amount, this throws `SwapAboveLimitError` (uncaught by the probe call) and aborts a max-swap that should have succeeded once fees are subtracted.

Note `letsexchange.ts` has the identical pattern, so this is an inherited gap in how `getMaxSwappable` is used across plugins, not unique to this PR — but it's directly reachable through this new code and worth a follow-up fix (e.g. skip/relax the max check during the probe, or have `getMaxSwappable` clamp the probe amount against known limits first).

src/swap/central/nym.ts
```diff
@@ -0,0 +1,378 @@
...
+  }
+
+  /**
+   * Maps a non-OK NYM response to the appropriate swap error. Bodies are parsed
+   * when present; otherwise classification falls back to the status code.
+   */
+  const handleErrorResponse = async (
+    response: { status: number; text: () => Promise<string> },
+    request: EdgeSwapRequestPlugin,
+    stage: 'quote' | 'order'
```
j0ntz, line 157: **PLAUSIBLE — correctness.** `handleErrorResponse` only special-cases geo-restriction text and 403s; any other order-stage rejection (e.g. the quoted amount falling out of range between the `/quote` and `/order` calls, which this two-step flow makes structurally possible) falls through to a generic `SwapCurrencyError`/untyped `Error` instead of a typed `SwapBelowLimitError`/`SwapAboveLimitError`. Sibling plugins (exolix's 422 handling, sideshift's `checkQuoteError`) map amount-out-of-range responses to typed limit errors so the UI can show the correct bound. Worth considering the same here for the order-stage error path.

src/swap/central/nym.ts
```diff
@@ -0,0 +1,378 @@
...
+// (`thorchain`) differs from the `thorchainrune` currency plugin.
+const pluginId = 'nymswap'
+
+// Edge pluginId of the NYM *asset/currency*. NYM provides its own liquidity, so
+// every swap must have the NYM asset on one side (enforced below).
+const NYM_PLUGIN_ID = 'nym'
+
+// Chains whose deposit/payout address must be a specific address type. Zcash
+// swaps require the transparent address rather than the default (shielded).
+const addressTypeMap: StringMap = {
```
j0ntz: **CONFIRMED — dead code.** `src/mappings/nym.ts` only maps `bitcoin`/`ethereum`/`nym`; `checkWhitelistedMainnetCodes` (called before `fetchSwapQuoteInner` runs) rejects any zcash wallet before this lookup could ever fire. Looks like a leftover from copy-pasting a sibling plugin's `addressTypeMap` — harmless today, but may mislead a future reader into thinking zcash is supported. Consider removing until zcash is actually added to the mapping.

src/swap/central/nym.ts
```diff
@@ -0,0 +1,378 @@
...
+   */
+  const handleErrorResponse = async (
+    response: { status: number; text: () => Promise<string> },
+    request: EdgeSwapRequestPlugin,
+    stage: 'quote' | 'order'
+  ): Promise<never> => {
+    const text = await response.text().catch(() => '')
+    log.warn(`NYM ${stage} error ${response.status}: ${text}`)
+
+    let parsed: unknown
```
j0ntz, line 162: **Cleanup / reuse.** Three patterns here are each reimplemented for the fourth time in this codebase with no shared helper: (a) try/`JSON.parse`-catch-undefined (lines 167-172, cf. `nexchange.ts:218-220`), (b) regex-based geo-restriction sniffing (line 180 — note this regex is broader than sibling plugins' equivalents, e.g. `sideshift.ts`'s `/country-blocked/i`, raising false-positive risk), and (c) the generic `"<Provider> returned error code <status>: <text>"` throw (lines 192-198, cf. `nexchange.ts`/`template.ts`). Might be worth extracting a shared `swapHelpers.ts` utility so future plugins (and these) converge on one implementation.

src/swap/central/nym.ts
```diff
@@ -0,0 +1,378 @@
...
+      await handleErrorResponse(quoteResponse, request, 'quote')
+    }
+
+    const quote = asNymQuote(await quoteResponse.json())
+
+    // Enforce the min/max limits returned by the quote against the user's
+    // requested amount (NYM returns native-unit limits for both sides).
+    // Compare `request.nativeAmount`, not the quote's echoed amount, so a
+    // clamped quote that omits an error still trips the limit error.
+    if (quoteFor === 'to') {
```
j0ntz: **Cleanup / simplification.** Lines 258-272 and 273-280 are structurally identical (lt-check → `SwapBelowLimitError`, gt-check → `SwapAboveLimitError`), differing only in which quote fields and direction label are used. Could collapse to picking a `{ min, max, side }` triple once based on `quoteFor`, then running one shared check block — removes the duplication and the risk of the two branches drifting apart on a future edit.

src/swap/central/nym.ts
```diff
@@ -0,0 +1,378 @@
...
+      }`
+    )
+  }
+
+  const fetchSwapQuoteInner = async (
+    request: EdgeSwapRequestPlugin
+  ): Promise<SwapOrder> => {
+    const { fromWallet, toWallet, quoteFor } = request
+
+    // NYM provides its own liquidity: one side of the swap must be NYM. Gate
```
j0ntz, line 217: **Altitude.** `swapHelpers.ts` already generalizes "this asset is disallowed on this side of a swap" via `InvalidTokenIds`/`checkInvalidTokenIds`. NYM's requirement ("NYM must be present on one side") is the mirror case, but is implemented here as a bespoke two-line pluginId check that bypasses that shared mechanism entirely (and, per the self-swap comment above, misses the `isSameAsset` guard that comes bundled with it). If another liquidity-provider-only plugin is added later, it'll likely reinvent this same check a third time. Worth considering a generalized "required asset" complement to `InvalidTokenIds` in `swapHelpers.ts`.

j0ntz (commented):
Automated review: 3 confirmed correctness bugs (most notably a double-order-creation issue on max-swap requests), 1 plausible correctness gap, and 4 cleanup/altitude items.

src/swap/central/nym.ts
j0ntz: Fixed. Split quote-fetching from order-creation. `fetchQuote` now performs the quote-only step (no order). `getMaxSwappable` probes via the new `fetchProbeOrder`, which builds the getMaxSpendable spendInfo from a quote alone, targeting the user's own from-chain refund address, so no live order is created during the probe. Only `fetchSwapQuoteInner` creates the order (once), from the trimmed amount. A max-swap now creates exactly one order instead of two.

src/swap/central/nym.ts
j0ntz: Fixed. Added `checkInvalidTokenIds({ from: {}, to: {} }, request, swapInfo)` in `fetchSwapQuote`, alongside `checkWhitelistedMainnetCodes`. It carries the shared `isSameAsset` guard, so a native-NYM self-swap (fromWallet === toWallet, same tokenId) is now rejected client-side before any NYM endpoint is hit, matching the sibling central plugins.

src/swap/central/nym.ts
j0ntz: Fixed. `fetchQuote` now takes an `enforceMax` flag. The `getMaxSwappable` probe (`fetchProbeOrder`) passes `false`, so an above-limit raw pre-fee balance clamps via getMaxSpendable instead of throwing `SwapAboveLimitError` and aborting the max-swap. The real `fetchSwapQuoteInner` passes `true`, so a genuinely above-max fee-adjusted amount still errors.

src/swap/central/nym.ts
j0ntz: Addressed by the two-step split. Limit enforcement now runs at the quote stage in `fetchQuote` (before the order call), and the order is created from the server-stored `quoteId`, so the quoted amount cannot drift between /quote and /order and range rejections do not surface at the order stage in this flow. The generic order-stage handler covers genuinely transient order failures. If NYM later returns typed amount-range errors at the order stage, mapping them to SwapBelow/AboveLimitError is a small follow-up.

src/swap/central/nym.ts
j0ntz: Fixed. Removed the `addressTypeMap` (its only entry was the dead zcash `transparentAddress` mapping, unreachable since `checkWhitelistedMainnetCodes` rejects any wallet not in the bitcoin/ethereum/nym mapping). Address lookups now use the default address type via `getAddress(wallet)`, and the now-unused `StringMap` import was dropped.

src/swap/central/nym.ts
j0ntz: Deferred as a cross-plugin follow-up. Extracting shared `swapHelpers` utilities for the try/JSON.parse-catch, geo-restriction sniffing, and generic error throw would touch nexchange/template/sideshift and the other central plugins, beyond this PR's scope. The broader geo regex (region|geo|country|blocked|jurisdiction) is intentional for NYM's stated geo policy rather than a copy of a narrower sibling. Tracked for a shared-helper consolidation pass.

src/swap/central/nym.ts
j0ntz: Fixed. Collapsed the two limit-check branches into a single `[min, max, side]` tuple selected once by `quoteFor`, followed by one below-limit and one above-limit check. Removes the duplicated structure and the drift risk.

src/swap/central/nym.ts
j0ntz: Deferred as a framework follow-up. Generalizing a 'required asset on one side' complement to `InvalidTokenIds` in `swapHelpers` is a framework addition for future liquidity-provider-only plugins, beyond this PR. The concrete self-swap gap this referenced is now closed by calling `checkInvalidTokenIds` (with its `isSameAsset` guard) in `fetchSwapQuote`.

j0ntz:
## 📸 Test evidence (pr-address retest): live mainnet NYM swap via updated plugin

**agent proof 1215306512581867 02 nym quote**

agent proof 1215306512581867 02 nym quote

**agent proof 1215306512581867 03 nym swap success**

agent proof 1215306512581867 03 nym swap success

**agent proof 1215306512581867 04 nym tx details**

agent proof 1215306512581867 04 nym tx details

peachbits (approved):

PR #459, opened 2026-06-10
Add transparent refund address for shielded Zcash Maya swaps

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

Asana: https://app.asana.com/0/1215088146871429/1214541361005860

**Problem.** When Edge sends a shielded-Zcash swap to Maya, the swap memo carries no refund address. Maya cannot refund a shielded source and rejects the refund:

```
can't refund ZEC from shielded source without a refund address
fail to refund for halted trading
```

For every chain except Zcash, Maya defaults the refund to the sending address, which is correct. A shielded ZEC source has no transparent sending address to refund to, so the integrator must supply a transparent (t-address) refund address in the swap memo.

**Fix.** In the shared Thorchain/Maya common code, when the source wallet is Zcash, fetch the wallet's transparent address (`getAddress(fromWallet, 'transparentAddress')`, the same helper already used for the ZEC receive side) and append it to the swap memo's destination field as `DESTADDR/REFUNDADDR` per the Maya memo spec, via the new documented helper `appendMayaRefundAddress` (which replaces the destination address in the memo with `DESTADDR/REFUNDADDR`). Since this is the Zcash-only path, the code throws `SwapCurrencyError` if the transparent refund address is missing.

The refund address is deliberately NOT passed to the `quote/swap` endpoint. Given a `refund_address`, Maya builds that same `DESTADDR/REFUNDADDR` memo and then rejects the quote because the result overflows the source chain's memo limit (e.g. ZEC to DASH is 86/80). The shielded ZEC send instead carries the memo in the encrypted note (512 bytes), which is not bound by the 80-char transparent-memo limit that the quote endpoint pre-validates against, so the refund is injected client-side after the quote is fetched without it. Non-Zcash sources are untouched and keep defaulting to the sending address (their quote requests and memos are unchanged).

The change lives entirely in the dependency; no gui-side wiring is required.

**Verification.**
- `tsc --noEmit`, eslint, and the mocha suite all pass (via `verify-repo.sh`, which also runs the webpack build).
- New unit tests in `test/thorchain.test.ts` pin `appendMayaRefundAddress`'s output to the exact memo shape (`=:d:DEST/REFUND:0/1/1:ej:75`) and cover the destination-not-present edge case.
- Live Maya node (`mayanode.mayachain.info`) confirms why the endpoint cannot build the memo: BTC to ETH with `refund_address` returns `generated memo too long for source chain: ...(96/80)`; the same request without it returns a valid 53-char memo.
- In-app ZEC to asset swap-to-success was driven to the success scene in the prior run (with the byte-identical memo). It could not be re-driven this pass: Maya ZEC trading is halted again (`HALTZECTRADING=1`, `trading is halted, can't process swap`) at the protocol level, and a swap during the halt fails at the quote before the refund injection runs, so it would not exercise this code anyway. Re-run the in-app smoke once Maya resumes ZEC trading.

---

> [!NOTE]
> **Medium Risk**
> Changes Maya Zcash swap memo construction and refund routing for a subset of swaps; wrong address formatting could misroute refunds, but scope is limited to ZEC source via the existing ZIP-321 path.
> 
> **Overview**
> **Shielded Zcash → Maya swaps** now embed a **transparent (t-address) refund** in the swap memo as `DESTADDR/REFUNDADDR`, so Maya can refund when trading is halted or the swap fails—shielded sources cannot default to the sender like other chains.
> 
> Quotes still call `quote/swap` **without** `refund_address`, because Maya would build the same long memo and reject quotes that exceed the **80-character transparent memo limit**; the shielded ZIP-321 path carries the memo in the encrypted note (512 bytes), so `appendMayaRefundAddress` patches the memo **after** quoting. Missing transparent refund address fails the swap with `SwapCurrencyError`.
> 
> Unit tests lock the memo shape; CHANGELOG notes the Maya behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 80f0430aee9e4c51c33d5791890df026e9f10e3e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/swap/defi/thorchain/thorchainCommon.ts
+++ b/src/swap/defi/thorchain/thorchainCommon.ts
@@ -889,12 +889,16 @@ export function makeThorchainBasedPlugin(
         fromCurrencyCode
       )
 
+      // Maya cannot refund a shielded Zcash source without a transparent refund
+      // address. This is the Zcash-only path, so it must be present.
+      if (refundAddress == null || refundAddress === '') {
+        throw new SwapCurrencyError(swapInfo, request)
+      }
+
       // Inject the transparent refund address into the shielded ZEC swap memo,
       // which is fetched from the quote endpoint without it (see
       // `appendMayaRefundAddress`).
-      if (refundAddress != null) {
-        memo = appendMayaRefundAddress(memo, refundAddress)
-      }
+      memo = appendMayaRefundAddress(memo, toAddress, refundAddress)
 
       // Encode memo per ZIP-321 as base64url (unpadded).
       const memoBase64Url = base64urlnopad.encode(utf8.decode(memo))
@@ -1621,19 +1625,17 @@ export const getVolatilitySpread = ({
  * note (512 bytes), which is not bound by that limit, so the refund is appended
  * here after the quote is fetched without it.
  *
- * The memo shape is `=:ASSET:DESTADDR:...`, so the refund is appended to the
- * destination field (index 2). A memo with no destination field is returned
- * unchanged.
+ * Maya reads the refund from the memo's destination field as
+ * `DESTADDR/REFUNDADDR`, so the destination address in the memo is replaced
+ * with `DESTADDR/REFUNDADDR`. A memo that does not contain the destination
+ * address is returned unchanged.
  */
 export const appendMayaRefundAddress = (
   memo: string,
+  destinationAddress: string,
   refundAddress: string
-): string => {
-  const memoFields = memo.split(':')
-  if (memoFields.length <= 2) return memo
-  memoFields[2] = `${memoFields[2]}/${refundAddress}`
-  return memoFields.join(':')
-}
+): string =>
+  memo.replace(destinationAddress, `${destinationAddress}/${refundAddress}`)
 
 /**
  * This will return the expected amount out from the quote maintaining backwards
--- a/src/swap/defi/thorchain/thorchainCommon.ts
+++ b/src/swap/defi/thorchain/thorchainCommon.ts
@@ -451,6 +451,16 @@ export function makeThorchainBasedPlugin(
         : undefined
     )
 
+    // Maya cannot refund a shielded Zcash source without a transparent
+    // (t-address) refund address; other chains default to the sending address,
+    // which is correct, so no refund address is needed. The address is injected
+    // into the ZEC swap memo below via `appendMayaRefundAddress` (see that
+    // helper for why it is not passed to the quote endpoint).
+    const refundAddress =
+      fromWallet.currencyInfo.pluginId === 'zcash'
+        ? await getAddress(fromWallet, 'transparentAddress')
+        : undefined
+
     const fromMainnetCode =
       MAINNET_CODE_TRANSCRIPTION[fromWallet.currencyInfo.pluginId]
     const toMainnetCode =
@@ -879,18 +889,12 @@ export function makeThorchainBasedPlugin(
         fromCurrencyCode
       )
 
-      // Debug logging for Maya ZEC ZIP-321 construction
-      log(
-        '[MAYA ZEC] zip321 components ' +
-          JSON.stringify({
-            chain: fromCurrencyCode,
-            inbound_address: thorAddress,
-            memo_ascii: memo,
-            memo_recipient: memoRecipient,
-            amount_native: fromNativeAmount,
-            amount_decimal: amountZec
-          })
-      )
+      // Inject the transparent refund address into the shielded ZEC swap memo,
+      // which is fetched from the quote endpoint without it (see
+      // `appendMayaRefundAddress`).
+      if (refundAddress != null) {
+        memo = appendMayaRefundAddress(memo, refundAddress)
+      }
 
       // Encode memo per ZIP-321 as base64url (unpadded).
       const memoBase64Url = base64urlnopad.encode(utf8.decode(memo))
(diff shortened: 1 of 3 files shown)


src/swap/defi/thorchain/thorchainCommon.ts
```diff
@@ -846,6 +858,23 @@ export function makeThorchainBasedPlugin(
...
+      // memo in the shielded note (not the transparent memo), which is not bound
+      // by that limit, so inject the refund into the memo here, after the quote
+      // was fetched without it.
+      if (refundAddress != null) {
+        // Memo shape is `=:ASSET:DESTADDR:...`; append the refund to the
+        // destination field so it becomes `DESTADDR/REFUNDADDR`.
+        const memoFields = memo.split(':')
+        if (memoFields.length > 2) {
+          memoFields[2] = `${memoFields[2]}/${refundAddress}`
+          memo = memoFields.join(':')
```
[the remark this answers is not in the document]
j0ntz: Invalid (wrong about the memo format). The Maya/Thorchain swap memo is `=:ASSET:DESTADDR:...`; `"=:d:dest".split(":")` returns `["=", "d", "dest"]`, so index 0 is `"="` (not an empty segment), index 1 is the asset code, and index 2 is the destination. The refund is appended to index 2 (the destination), giving `=:d:DESTADDR/REFUNDADDR:...`.

Verified at runtime on a Metro-connected simulator (ZEC to DASH via Maya): the constructed shielded-note memo was `=:d:XnzErKGuqcG5Ci5oTsQv7stwCBofgChu8s/t1MnUkHpi3Ampr9ZzAtnWHFbatsVF3hEvKL:0/1/1:ej:75`, with the transparent refund t-address correctly on the destination field, byte-identical to the memo Maya itself returns when given `refund_address`. The swap then executed to its success scene.

j0ntz:
## 📸 Test evidence (in-app, after fix)

**agent proof 1214541361005860 04 zec dash fixed order**

agent proof 1214541361005860 04 zec dash fixed order

**agent proof 1214541361005860 05 zec dash swap success**

agent proof 1214541361005860 05 zec dash swap success

**agent proof 1214541361005860 06 btc eth unaffected order**

agent proof 1214541361005860 06 btc eth unaffected order

src/swap/defi/thorchain/thorchainCommon.ts
```diff
@@ -879,6 +858,23 @@ export function makeThorchainBasedPlugin(
...
       )
 
+      // Maya cannot refund a shielded Zcash source without a transparent refund
+      // address, carried in the swap memo as `DESTADDR/REFUNDADDR`. It cannot be
+      // sent to the quote endpoint: appending it overflows Zcash's 80-char
+      // transparent-memo limit and the quote is rejected. The ZEC swap sends its
+      // memo in the shielded note (not the transparent memo), which is not bound
+      // by that limit, so inject the refund into the memo here, after the quote
+      // was fetched without it.
+      if (refundAddress != null) {
```
peachbits: This blob is pretty ugly. The memo is created by the quote/swap endpoint and it accepts a refund address argument so let's pass it to taht and use the memo they return.

peachbits (changes requested):

src/swap/defi/thorchain/thorchainCommon.ts
j0ntz: Passing the refund address to the quote/swap endpoint is what the original PR did, and it breaks every ZEC swap. Given `refund_address`, Maya builds the same `DESTADDR/REFUNDADDR` memo and then rejects the quote because the result overflows the source chain's memo limit. Reproduced live today against mayanode:

- BTC->ETH with refund: `generated memo too long for source chain: =:e:0x742d...f44e/bc1q...5mdq::ej:75 (96/80)`
- ZEC->DASH with refund is 86/80 (captured live in the prior run's window when ZEC trading was open).

The reason it still works on-chain is that the shielded ZEC send carries the memo in the encrypted note (512 bytes), which is not bound by the 80-char transparent-memo limit that Maya's quote endpoint pre-validates against. So the endpoint cannot return a usable memo for a shielded ZEC source; the refund has to be appended client-side after the quote is fetched without it.

I've addressed the readability concern: the inline blob is now a single call to a named, documented helper `appendMayaRefundAddress` (src/swap/defi/thorchain/thorchainCommon.ts), with the endpoint/80-char rationale on its docstring, plus unit tests in test/thorchain.test.ts pinning the exact memo output. Leaving this thread open for your call on the approach.

src/swap/defi/thorchain/thorchainCommon.ts
```diff
@@ -879,6 +889,13 @@ export function makeThorchainBasedPlugin(
...
 
+      // Inject the transparent refund address into the shielded ZEC swap memo,
+      // which is fetched from the quote endpoint without it (see
+      // `appendMayaRefundAddress`).
+      if (refundAddress != null) {
+        memo = appendMayaRefundAddress(memo, refundAddress)
+      }
+
       // Debug logging for Maya ZEC ZIP-321 construction
       log(
```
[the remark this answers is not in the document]
j0ntz: Fixed: removed the debug `log()` that emitted the constructed memo (`memo_ascii`), along with the rest of that per-swap construction dump (Maya recipient, amounts). Nothing now writes the shielded-to-transparent linkage to the logs.

src/swap/defi/thorchain/thorchainCommon.ts
peachbits: It isn't more readable, the ugly blob was just moved to a different part of the file. That utility can be improved by replacing the destinationAddress part of the string with a destinationAddress/refundAddress string. 

`const withRefund = memo.replace(destination, `${destination}/${refundAddress}`)
 `

Also, since we're in a zcash-only code path, we should throw a swapcurrencyerror if the refund address isn't present (even if that is really unlikely to happen)

peachbits (changes requested):

src/swap/defi/thorchain/thorchainCommon.ts
j0ntz: Done, both points. `appendMayaRefundAddress` now does `memo.replace(destinationAddress, `${destinationAddress}/${refundAddress}`)` instead of splitting on `:` and indexing. And the Zcash-only path now throws `SwapCurrencyError` when the refund address is absent (`refundAddress == null || refundAddress === ''`) before the memo is built. Tests updated to the 3-arg signature and verify-repo is green (HEAD 80f0430).

PR #470, opened 2026-07-03
Buffer swap provider gas prices to avoid below-minimum race

peachbits:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

Swap partners (LI.FI, Rango, Unizen) return a gas price alongside their quote, which we pass straight through as a `customNetworkFee.gasPrice`. There's a race between that quote and when the engine actually signs the transaction: the account-based engine clamps the minimum acceptable gas price to the **current network base fee**, and on fast-moving chains the base fee can tick up in the gap. When it does, the provider's now-stale gas price lands below that floor and the send is rejected — e.g. on Arbitrum:

> Gas price 20084000 wei below minimum 20208000 wei

To give the quoted price enough headroom to clear the floor, apply a small **5% buffer** to the provider-supplied gas price via a shared `bufferSwapGasPrice` helper in `defiUtils.ts`, used at each provider's gas-price conversion point (LI.FI, Rango, Unizen).

Notes:
- The buffer is modest on purpose. The engine derives the EIP-1559 tip as `maxFeePerGas - baseFee`, so the buffered value is actually spent rather than being free ceiling headroom.
- Rango's EIP-1559 `maxFeePerGas` / `maxPriorityFeePerGas` path is left untouched — only the legacy `gasPrice` field (the one that hits the floor check) is buffered.
- Unizen unit-test fixtures updated to the buffered gas prices.

---

> [!NOTE]
> **Medium Risk**
> Changes real network fees on EVM DEX swaps (users pay the buffered gas), but scope is limited to fee calculation on three providers with a modest multiplier.
> 
> **Overview**
> Adds shared **`bufferSwapGasPrice`** in `defiUtils.ts` that bumps provider-quoted gas (in wei) by **5%**, rounds to integer wei, then returns gwei so `customNetworkFee.gasPrice` stays above the network minimum when base fee moves between quote and sign.
> 
> **LI.FI**, **Rango** (legacy `gasPrice` only; EIP-1559 `maxFeePerGas` / `maxPriorityFeePerGas` unchanged), and **Unizen** now use this helper instead of passing the raw quoted rate. Unizen **`makeEvmFees`** normalizes gwei inputs to wei before buffering.
> 
> Also bumps **`webpack-cli`** from ^4.6.0 to ^5.1.4 (lockfile) and updates Unizen unit-test expected gas prices.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a69c135ebcf7ce0806214a5279b76f1e46c7a548. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
  - https://app.asana.com/0/0/1216067423670379

### DIFF
--- a/src/swap/defi/defiUtils.ts
+++ b/src/swap/defi/defiUtils.ts
@@ -1,3 +1,4 @@
+import { div, mul, round } from 'biggystring'
 import { asMaybe, asObject, asString } from 'cleaners'
 import {
   EdgeCurrencyConfig,
@@ -233,3 +234,30 @@ export const getDepositWithExpiryData = async (params: {
 }
 
 export const WEI_MULTIPLIER = '1000000000'
+
+export const SWAP_GAS_PRICE_BUFFER_MULTIPLIER = '1.05'
+
+/**
+ * Swap providers quote a gas price at quote time, but the network base fee can
+ * rise before we sign the transaction, causing it to be rejected as below the
+ * network minimum (`Gas price X wei below minimum Y wei`). Apply a small buffer
+ * to the provider-supplied gas price to absorb that drift.
+ *
+ * The buffer is applied in wei and rounded to an integer wei value, so the
+ * returned gwei string never carries more than 9 decimal places. The engine
+ * multiplies `customNetworkFee.gasPrice` back to wei, and a fractional wei
+ * amount corrupts the transaction's hex encoding.
+ *
+ * Note the buffer is real spend, not free headroom: the engine derives the
+ * EIP-1559 tip as `maxFeePerGas - baseFee`, so the entire buffered value is
+ * paid. Keep the multiplier modest.
+ *
+ * @param gasPriceWei provider-quoted gas price, in wei
+ * @returns buffered gas price, in gwei
+ */
+export const bufferSwapGasPrice = (gasPriceWei: string): string =>
+  div(
+    round(mul(gasPriceWei, SWAP_GAS_PRICE_BUFFER_MULTIPLIER), 0),
+    WEI_MULTIPLIER,
+    18
+  )
--- a/src/swap/defi/lifi.ts
+++ b/src/swap/defi/lifi.ts
@@ -45,7 +45,10 @@ import {
   MakeTxParams,
   StringMap
 } from '../types'
-import { createEvmApprovalEdgeTransactions, WEI_MULTIPLIER } from './defiUtils'
+import {
+  bufferSwapGasPrice,
+  createEvmApprovalEdgeTransactions
+} from './defiUtils'
 
 const pluginId = 'lifi'
 const swapInfo: EdgeSwapInfo = {
@@ -448,7 +451,7 @@ export function makeLifiPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
         const transactionRequest = asTransactionRequest(transactionRequestRaw)
         const { data, gasLimit, gasPrice } = transactionRequest
         const gasPriceDecimal = hexToDecimal(gasPrice)
-        const gasPriceGwei = div18(gasPriceDecimal, WEI_MULTIPLIER)
+        const gasPriceGwei = bufferSwapGasPrice(gasPriceDecimal)
 
         if (sendingToken) {
           const approvalTxs = await createEvmApprovalEdgeTransactions({
--- a/src/swap/defi/rango.ts
+++ b/src/swap/defi/rango.ts
@@ -25,7 +25,6 @@ import {
 import { base64 } from 'rfc4648'
 
 import { rango as rangoMapping } from '../../mappings/rango'
-import { div18 } from '../../util/biggystringplus'
 import {
   getMaxSwappable,
   makeSwapPluginQuote,
@@ -50,9 +49,9 @@ import {
   StringMap
 } from '../types'
 import {
+  bufferSwapGasPrice,
   createEvmApprovalEdgeTransactions,
-  decodeEvmApprovalData,
-  WEI_MULTIPLIER
+  decodeEvmApprovalData
 } from './defiUtils'
 
 const swapInfo: EdgeSwapInfo = {
@@ -739,7 +738,7 @@ export function makeRangoPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
               : undefined,
           gasPrice:
             evmTransaction.gasPrice != null
-              ? div18(evmTransaction.gasPrice, WEI_MULTIPLIER)
+              ? bufferSwapGasPrice(evmTransaction.gasPrice)
               : undefined,
           maxFeePerGas: evmTransaction.maxFeePerGas ?? undefined,
           maxPriorityFeePerGas: evmTransaction.maxPriorityFeePerGas ?? undefined
--- a/src/swap/defi/unizen.ts
+++ b/src/swap/defi/unizen.ts
@@ -1,5 +1,5 @@
 import contracts from '@unizen-io/unizen-contract-addresses/production.json'
-import { div } from 'biggystring'
+import { mul } from 'biggystring'
 import {
   asArray,
   asMaybe,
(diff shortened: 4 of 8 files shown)


src/swap/defi/lifi.ts
```diff
@@ -448,7 +452,9 @@ export function makeLifiPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
         const transactionRequest = asTransactionRequest(transactionRequestRaw)
         const { data, gasLimit, gasPrice } = transactionRequest
         const gasPriceDecimal = hexToDecimal(gasPrice)
-        const gasPriceGwei = div18(gasPriceDecimal, WEI_MULTIPLIER)
+        const gasPriceGwei = bufferSwapGasPrice(
+          div18(gasPriceDecimal, WEI_MULTIPLIER)
```
j0ntz: **Bug: buffering a full-precision `div18` gwei value produces a non-integer wei amount, which crashes transaction signing.**

`div18(gasPriceDecimal, WEI_MULTIPLIER)` preserves up to 18 fractional gwei digits (e.g. `20.083948273`). `bufferSwapGasPrice` then multiplies by `1.05` via `biggystring.mul`, which does exact decimal multiplication with no rounding — so the result can carry up to ~11 decimal digits (e.g. `21.08814568665`).

Downstream, `edge-currency-accountbased`'s `calcMiningFees` converts this gwei value back to wei via `mul(gasPrice, '1000000000')`. Since the buffered value has more than 9 fractional digits, the resulting wei amount is itself fractional (e.g. `21088145686.65`). That value is later hex-encoded via `mul('1', gasPrice, 16)` in `getFeeParamsByTransactionType`, which happily emits a hex string with a literal `.` in it (e.g. `0xbca2ccb.d64c`) instead of throwing.

That malformed hex then reaches `@ethereumjs/tx`'s `TransactionFactory.fromTxData`, which throws:
```
Cannot convert string to Uint8Array. toBytes only supports 0x-prefixed hex strings and this string was given: 0xbca2ccb.d64c
```
Verified directly with a node repro reproducing this exact chain end-to-end.

Any lifi swap whose quoted wei gas price isn't a clean multiple of `10^9` (i.e. almost all of them) will hard-crash `makeSpend` after this buffer is applied, rather than merely risking the 'below minimum' rejection this PR is meant to fix. Suggest rounding/truncating the buffered gwei value to a fixed precision (e.g. 9 decimal places, matching wei granularity) before returning it from `bufferSwapGasPrice`, similar to how the adjacent `gasLimit` computation on line 485 wraps in `round(mul(...), 0)`.

src/swap/defi/rango.ts
```diff
@@ -739,7 +740,9 @@ export function makeRangoPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
               : undefined,
           gasPrice:
             evmTransaction.gasPrice != null
-              ? div18(evmTransaction.gasPrice, WEI_MULTIPLIER)
+              ? bufferSwapGasPrice(
+                  div18(evmTransaction.gasPrice, WEI_MULTIPLIER)
```
j0ntz: Same fractional-wei-precision bug as flagged on `lifi.ts` — `div18(evmTransaction.gasPrice, WEI_MULTIPLIER)` can carry up to 18 fractional gwei digits, and `bufferSwapGasPrice`'s `mul(..., '1.05')` extends that further with no rounding. When converted back to wei downstream (`mul(gasPrice, '1000000000')` in `edge-currency-accountbased`), a non-multiple-of-1e9 wei gas price produces a fractional wei value that later corrupts hex encoding and crashes `TransactionFactory.fromTxData` (`Cannot convert string to Uint8Array... 0xbca2ccb.d64c`). Same fix suggestion applies here: round/truncate to a fixed precision before returning from `bufferSwapGasPrice`.

src/swap/defi/rango.ts
```diff
@@ -739,7 +740,9 @@ export function makeRangoPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
...
+                )
               : undefined,
           maxFeePerGas: evmTransaction.maxFeePerGas ?? undefined,
```
j0ntz, line 743: `maxFeePerGas`/`maxPriorityFeePerGas` are passed through unbuffered while the legacy `gasPrice` field gets the 5% buffer. `calcMiningFees` in `edge-currency-accountbased`'s `ethMiningFees.ts` only reads `customNetworkFee.gasPrice`/`gasLimit`, not `maxFeePerGas`/`maxPriorityFeePerGas` from `customNetworkFee` — so for an EIP-1559 chain where Rango returns `maxFeePerGas`/`maxPriorityFeePerGas` instead of (or in addition to) legacy `gasPrice`, this PR's fix never applies at all, and the original 'gas price below minimum' race this PR targets remains possible for those swaps.

src/swap/defi/unizen.ts
```diff
@@ -65,7 +69,7 @@ const makeEvmFees = (rate: string, units: string = 'wei'): SwapNetworkFees => {
   const multiplier = units === 'wei' ? WEI_MULTIPLIER : 1
   return {
     customNetworkFee: {
-      gasPrice: div(rate, multiplier)
+      gasPrice: bufferSwapGasPrice(div(rate, multiplier))
```
j0ntz: `div(rate, multiplier)` here has no precision argument, which defaults to 0 (truncates to an integer gwei value) before `bufferSwapGasPrice` applies the 1.05 multiplier — unlike `lifi.ts`/`rango.ts`, which use the 18-decimal `div18` wrapper. For a gas price like `1999999999` wei (just under 2 gwei), this truncates to `'1'` gwei and buffers to `'1.05'`, silently discarding ~1 gwei of precision. On chains where sub-gwei precision matters, the nominal '5% buffer' can land well below the actual quoted rate, potentially reintroducing the exact 'below minimum' failure this PR is meant to fix. Test fixtures only exercise gas prices that divide evenly by `WEI_MULTIPLIER` (e.g. `1000000000`, `7832997367` truncating cleanly to `7`), so this truncation-then-buffer precision loss isn't caught by the added tests.

src/swap/defi/defiUtils.ts
```diff
@@ -233,3 +234,19 @@ export const getDepositWithExpiryData = async (params: {
...
+ * network minimum (`Gas price X wei below minimum Y wei`). Apply a small buffer
+ * to the provider-supplied gas price to absorb that drift.
+ *
+ * Note the buffer is real spend, not free headroom: the engine derives the
+ * EIP-1559 tip as `maxFeePerGas - baseFee`, so the entire buffered value is
+ * paid. Keep the multiplier modest.
+ *
+ * @param gasPriceGwei provider-quoted gas price, in gwei
+ */
+export const SWAP_GAS_PRICE_BUFFER_MULTIPLIER = '1.05'
```
j0ntz: Minor: this JSDoc block (with `@param gasPriceGwei`) sits directly above `SWAP_GAS_PRICE_BUFFER_MULTIPLIER`, a plain constant with no parameters, rather than above the `bufferSwapGasPrice` function it actually documents. Consider moving the comment down one declaration, or dropping the `@param` tag from the constant.

j0ntz:
One more finding from review, on code this PR doesn't touch but that shares the exact race condition being fixed here:

**`src/swap/defi/uni-v2-based/uniV2Utils.ts:123` and `src/swap/defi/uni-v2-based/plugins/velodrome.ts:172`** — `getSwapTransactions()` accepts a `previousGasPrice` (a gas price cached from an earlier quote via `customFeeCache`) and uses it verbatim via `ethers.utils.parseUnits`, with no buffer applied. The uni-v2-based plugins (spookyswap, tombswap, velodrome, etc.) reuse a quote-time gas price for the actual signed swap transaction the same way lifi/rango/unizen do, so on fast-moving chains they're exposed to the identical 'Gas price below minimum' race this PR fixes for the other three providers, but `bufferSwapGasPrice` was never wired into this shared helper. Worth a follow-up PR to close that gap, if the goal is to eliminate this failure mode across all EVM swap plugins rather than just three of them.

j0ntz (approved):

PR #471, opened 2026-07-10
Add Changelly swap plugin

peachbits:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

Adds Changelly as a centralized swap provider, superseding #449.

This re-cuts changelly/edge-exchange-plugins#feat_changelly_integration (head 86b41f4) onto current master as a fresh commit — master has moved substantially since that branch was cut (npm conversion in 2.47.0, several releases), so a re-cut resolved the conflicts more cleanly than another rebase. Changelly's authorship is preserved via a `Co-authored-by` trailer.

GUI half: EdgeApp/edge-react-gui#6028 supplies `CHANGELLY_INIT` (XOR-obfuscated apiKey plus plaintext partnerId, both required, no defaults).

**Changes vs. the #449 head**

- **`partnerId` initOption**: the JSON-RPC request id now carries `initOptions.partnerId` (Edge passes `edge-app`) instead of a hardcoded literal, so the partner identifier is configuration rather than plugin source.
- **Required-credentials guard**: the factory throws when `initOptions.apiKey` or `initOptions.partnerId` is `''` (a missing member already throws in the `asInitOptions` cleaner). edge-core-js catches factory throws per-plugin, so an unconfigured build simply runs without Changelly — no quotes, no crash, and no default values anywhere.
- **Limit-amount unit fix**: min/max limits from `Invalid amount` errors are now converted to native units *before* integer truncation. Previously the denominated value was truncated first, so a sub-unit limit (e.g. 0.0025 BTC) collapsed to a native amount of "0" in `SwapBelowLimitError`/`SwapAboveLimitError`.
- **Conventions**: named `makeChangellyPlugin` export (was an `export default` arrow), sorted imports, lint/prettier normalization.
- **Test harness**: `CHANGELLY_INIT` wired into `test/testconfig.ts` / `test/testpartner.ts` for manual live-quote testing (separate commit, droppable).

**Notes**

- Auth otherwise matches the #449 head: `X-Auth: base64(apiKey:timestamp)` against Changelly's relay endpoint. No partner secret ships in the app.
- `src/mappings/changelly.ts` / `scripts/mappings/changellyMappings.ts` are mapctl-format seed files; there is no changelly synchronizer in `scripts/allSynchronizers.ts` yet (future work — mappings update by hand until then).

**Verification**

- `npm run lint`, `npm run types`, `npm test` clean.
- Factory contract exercised directly: missing/empty apiKey or partnerId throws; valid values construct the plugin.
- Manual live quote via the GUI (`DEBUG_EXCHANGES` against `npm start`): fixed-rate quotes returned and swaps completed against the relay with real credentials.
- Negative paths: no `CHANGELLY_INIT` → plugin skipped by the core; empty apiKey/partnerId → factory throws, core logs and continues.

---

> [!NOTE]
> **Medium Risk**
> New third-party swap path handles user funds (pay-in addresses, amounts, memos) and partner credentials; behavior is gated on init options but live swaps depend on correct API/limit handling.
> 
> **Overview**
> **Re-introduces Changelly** as a centralized swap provider (replacing the long-deprecated plugin) and registers it on the core plugin bundle.
> 
> The new **`makeChangellyPlugin`** talks to Changelly’s relay over JSON-RPC with **`X-Auth`** (`apiKey` + timestamp) and requires **`apiKey`** and **`partnerId`** in init options—the factory throws if either is empty so unconfigured builds skip Changelly without crashing. **`partnerId`** is also used in the JSON-RPC request id instead of a hardcoded value.
> 
> Quotes use **fixed-rate** flow: **`getFixRateForAmount`** then **`createFixTransaction`**, with dynamic asset discovery from **`getCurrenciesFull`** and chain/ticker resolution via new **Changelly ↔ Edge plugin mappings** (seed + generated `src/mappings/changelly.ts`). Min/max errors map **`error.data.limits`** to **`SwapBelowLimitError`** / **`SwapAboveLimitError`** with limits converted to **native units before** integer truncation. Zcash uses transparent addresses.
> 
> Test harness adds **`CHANGELLY_INIT`**; CHANGELOG notes the new provider.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 957fa08fbbd0fa0bd102f4a8be81c962dbe18cbf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
  - https://app.asana.com/0/0/1213981191982796


src/swap/central/changelly.ts
```diff
@@ -0,0 +1,611 @@
...
+  fetch: EdgeFetchFunction,
+  apiKey: string,
+  partnerId: string
+): ChangellyClient {
+  const changellyClientRequest = async <
+    T extends Object | undefined,
+    R = any
+  >(
+    body: Omit<Body<T>, 'jsonrpc' | 'id'>,
+    promoCode?: string
```
j0ntz: The `promoCode` param is declared here but never used inside `changellyClientRequest` (not added to `jsonBody` or `headers`). `fetchSwapQuote`'s `opts.promoCode` (line 587) is threaded through `fetchSwapQuoteInner` but that function never reads `opts` at all, and the only call site that passes this param is `getCurrenciesFull`, hardcoded to `undefined`. Any promo code a caller supplies is silently dropped before it reaches the API. Either wire it into the request or drop the dead parameter/opts plumbing.

src/swap/central/changelly.ts
```diff
@@ -0,0 +1,611 @@
...
+  ): Promise<Result<R> | ErrorResult> => {
+    const jsonBody = JSON.stringify({
+      ...body,
+      jsonrpc: '2.0',
+      id: body.method + ':' + partnerId
+    })
+
+    const headers: Record<string, string> = {
+      'Content-Type': 'application/json',
+      'X-Auth': btoa([apiKey, Date.now()].join(':'))
```
j0ntz: `btoa` is used here to build the X-Auth header, but nothing else in this repo uses `btoa` (base64 encoding elsewhere goes through `Buffer.from(...).toString('base64')` or the `rfc4648` package, e.g. src/swap/defi/rango.ts, src/swap/defi/lifi.ts). If the runtime this plugin executes in doesn't expose `btoa` as a global, every request throws `ReferenceError: btoa is not defined` and the plugin never works. Worth confirming `btoa` is actually available in the target runtime, or switching to the `Buffer`/`rfc4648` pattern used everywhere else.

src/swap/central/changelly.ts
```diff
@@ -0,0 +1,611 @@
...
+      fixRateParams.amountTo = quoteAmount
+    } else {
+      fixRateParams.amountFrom = quoteAmount
+    }
+
+    const rateResponse = await client.getFixRateForAmount(fixRateParams)
+
+    if (isError(rateResponse)) {
+      const { message: msg, data: errorData } = rateResponse.error
+      if (msg.startsWith('Invalid amount')) {
```
j0ntz: The below/above-limit branch decides direction by matching English substrings in the error message (`msg.includes('Minimal')` / `msg.includes('Maximum')`) even though `errorData.limits` already carries both `min` and `max` numerically in the same payload. If Changelly changes the wording or capitalization of this message, both branches silently stop matching and execution falls through to a generic `SwapCurrencyError`, losing the specific min/max amount shown to the user. Sibling plugins (changehero.ts, changenow.ts) determine below/above-limit by comparing quote amounts numerically against min/max fields instead of parsing message text. Worth doing the same here rather than relying on message wording.

src/swap/central/changelly.ts
```diff
@@ -0,0 +1,611 @@
...
+    displayName: 'Changelly',
+    supportEmail: 'support@changelly.com'
+  }
+
+  const fetchSupportedAssets = async (): Promise<void> => {
+    if (lastUpdated > Date.now() - EXPIRATION) return
+
+    const data = await client.getCurrenciesFull()
+
+    if (isError(data) && data.error.code === -32600) {
```
j0ntz: Magic number `-32600` with no comment or named constant. It's the JSON-RPC 2.0 standard 'Invalid Request' code, not obviously verification-related, so nothing here documents why this specific code maps to `SwapPermissionError('noVerification')`. A named constant or a short comment would save the next person from having to reverse-engineer this against Changelly's docs.

src/swap/central/changelly.ts
```diff
@@ -0,0 +1,611 @@
...
+      }
+
+      chainCodeTickerMap = out
+      lastUpdated = Date.now()
+    } catch (e: unknown) {
+      log.warn('Changelly: Error updating supported assets', e)
+    }
+  }
+
+  fetchSupportedAssets().catch((e) => {
```
j0ntz: This fires an unawaited `fetchSupportedAssets()` at plugin construction, in addition to the awaited call inside `fetchSwapQuote` (line 591) which already guarantees the map is populated before use given the internal 1-hour cache check. Sibling plugins (changehero.ts, changenow.ts) only call their equivalent lazily inside `fetchSwapQuote`, not at construction. This adds an extra network request on every plugin instantiation regardless of whether the user ever opens Changelly, and can race with an early `fetchSwapQuote` call into two concurrent `getCurrenciesFull` requests.

src/swap/central/changelly.ts
```diff
@@ -0,0 +1,611 @@
...
+          request.nativeAmount,
+          request.toTokenId
+        )
+      : nativeToDenomination(
+          request.fromWallet,
+          request.nativeAmount,
+          request.fromTokenId
+        )
+
+    const fixRateParams: any = {
```
j0ntz: `fixRateParams` is typed `any`, which bypasses the `FixEstimationRequest` union's amountFrom/amountTo check. The same reverseQuote-conditional amountFrom/amountTo assignment is duplicated below for `txParams` (lines 501-505). A small helper (e.g. `applyQuoteAmount(params, reverseQuote, quoteAmount)`) would remove the duplication and let `fixRateParams` keep the `FixEstimationRequest` type instead of `any`.

src/swap/central/changelly.ts
peachbits: Removed the construction-time call; the awaited fetchSupportedAssets inside fetchSwapQuote (with the 1-hour cache) is now the only fetch path, matching changehero/changenow.

src/swap/central/changelly.ts
```diff
@@ -0,0 +1,611 @@
...
+      chainCodeTickerMap = out
+      lastUpdated = Date.now()
+    } catch (e: unknown) {
+      log.warn('Changelly: Error updating supported assets', e)
+    }
+  }
+
+  fetchSupportedAssets().catch((e) => {
+    log.warn('Changelly: Error refreshing cache', e)
+  })
```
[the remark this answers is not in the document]
peachbits: The startup fetch block was removed entirely (per the eager-fetch thread), so this catch handler no longer exists.

src/swap/central/changelly.ts
peachbits: Dropped the dead promoCode parameter and the opts plumbing through fetchSwapQuoteInner/getMaxSwappable; getCurrenciesFull no longer passes undefined.

src/swap/central/changelly.ts
peachbits: Switched to base64.stringify from rfc4648 (the pattern lifi/rango already use), so the header no longer depends on a btoa global.

src/swap/central/changelly.ts
peachbits: Replaced the message matching with numeric comparison: when error.data.limits is present, quoteAmount is compared against the reverse-aware min/max via biggystring lt/gt, so wording changes can no longer break limit detection.

src/swap/central/changelly.ts
peachbits: Named it JSON_RPC_INVALID_REQUEST with a comment noting the relay returns it when it rejects the X-Auth partner credentials, hence the permission-error mapping.

src/swap/central/changelly.ts
peachbits: fixRateParams is now typed FixEstimationRequest (base type corrected to Omit amountFrom so the union works) and both call sites share a single quoteAmountParam object, removing the any and the duplicated conditional.

src/swap/central/changelly.ts
```diff
@@ -0,0 +1,585 @@
...
+
+      checkInvalidTokenIds(INVALID_TOKEN_IDS, request, swapInfo)
+      checkWhitelistedMainnetCodes(
+        MAINNET_CODE_TRANSCRIPTION,
+        request,
+        swapInfo
+      )
+
+      const newRequest = await getMaxSwappable(fetchSwapQuoteInner, request)
+      const swapOrder = await fetchSwapQuoteInner(newRequest)
```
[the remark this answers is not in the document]
peachbits: This is the repo-wide getMaxSwappable contract rather than a changelly-specific defect: the probe exists to obtain the partner payin address (swapOrder.spendInfo) for getMaxSpendable, and for every centralized provider that address only exists after order creation — changenow and changehero create their probe order the same way. Their quote paths likewise throw SwapAboveLimitError when balance exceeds the provider max (changehero.ts:257), which getMaxSwappable intentionally does not catch: the error carries the provider max for the UI. Avoiding throwaway probe orders would be a swapHelpers-level redesign affecting all centralized plugins, out of scope for this integration.

j0ntz (approved):

PR #18, opened 2019-05-22
Coinswitch plugin

ramendra11:
coinswitch plugin

PR #475, opened 2026-07-15
Add Swapter swap provider

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

Asana: https://app.asana.com/1/9976422036640/project/1213880789473005/task/1216571782597915

Adds **Swapter** as a centralized swap provider.

This recreates partner PR #455 (by @markovo4, from `markovo4/edge-exchange-plugins`) on an
EdgeApp branch so it can run CI, rebased from its `v2.46.0` merge-base onto current
`master`. The plugin design, mapping and test cases started as the partner's work; this
branch brings the integration up to current conventions, moves it onto Swapter's
Edge-specific adapter routes, and reconciles the mapping against the live asset list.

#### How the plugin works

**Endpoints.** Quotes use Swapter's Edge adapter routes,
`POST /adapter/edge/swap/deposit-range` and `POST /adapter/edge/swap/create`, rather than
the `/v2` originals. That is not a cosmetic swap. The adapter routes serialize every
numeric field as a string, where `/v2` returns unquoted JSON numbers that lose their low
digits in `JSON.parse` (an 18-decimal deposit round-trips exactly on the adapter route and
does not on `/v2`). And `deposit-range` is authoritative where `/v2/swap/min-amount` was
not: `min-amount` reported a floor BELOW the one create actually enforces, so an amount
could clear the plugin's own check and then be rejected by create.

**Rate type.** The plugin asks for a fixed-rate order first and falls back to a floating
one, which is the convention the sibling central plugins follow. Swapter spells the fixed type
`fix`, not `fixed`; an unrecognized value is not rejected as a bad request but answered
with HTTP 500, which is indistinguishable from a pair Swapter genuinely cannot fix, so the
wrong spelling silently downgraded every quote to float. The fallback still runs for pairs
that really cannot be fixed.

**Bounds and errors.** `deposit-range` supplies both a floor and a ceiling, and both round
INWARD into native units (minimums up, maximums down) so rounding can never widen the range
past what Swapter accepts. Create's own limit codes (`factory:6` below, `factory:9` above)
map to `SwapBelowLimitError` / `SwapAboveLimitError` whether or not Swapter echoes the
bound it rejected against. Unsupported-pair codes differ between the two routes, so both
vocabularies are recognised: `deposit-range` answers a bare service code, create namespaces
its own (`factory:1`/`2`/`3`). Errors are matched on `name` rather than `instanceof`,
because edge-core-js's error constructors return a plain `Error` carrying a `name` and
`instanceof` against them is permanently false.

**Max swaps.** The quote routes through `getMaxSwappable` like every other central plugin.
The fee probe targets the user's own refund address with `skipChecks: true`, so pricing a
max swap creates no abandoned order and EVM wallets do not fail with `SpendToSelfError`.
Exactly one order is created per quote.

**Assets.** `src/mappings/swapter.ts` maps each Edge plugin ID to a Swapter network name,
and `/data/coins` (fetched unauthenticated, since that route answers 401 for any
`X-API-KEY`) supplies the per-network ticker set used to resolve tokens. Every entry is
verified against the live list rather than reasoned about: a stale code is caught by no type
and no test, and simply fails at quote time. Three chains need a `SPECIAL_MAINNET_CASES`
override because Swapter's asset code differs from Edge's `currencyCode`: native TON is
`GRAM`, Bitcoin SV is `BCHSV`, and the EOS network carries only Vaulta's `A`. zkSync is
mapped for its `ZK` token but its native ETH is blocked through `INVALID_TOKEN_IDS`, since
Swapter does not list it. An empty ticker map is treated as a provider outage and fails the
quote, rather than being cached for the hour and reported as an unsupported pair.

`test/partnerJson/swapterMap.json` is regenerated from a current `/data/coins` snapshot
(44 networks, 829 tickers).

#### Known limitations

These are gaps against `API_REQUIREMENTS.md` that the plugin cannot close on its own,
because they need changes to Swapter's adapter. They are recorded here rather than worked
around:

- **Assets are identified by ticker plus network name**, not by chain plus contract. The
  adapter accepts only `coin` and `network`, so the plugin has no contract field to send
  even though it resolves contracts locally to build the ticker map.
- **No numeric EVM `chainId`.** EVM chains are keyed by Swapter's own network names, so
  adding a chain (or surviving a rename) needs a mapping edit.
- **Quotes are deposit-amount only**, so exact-receive (`to`) quotes are unsupported and
  `fetchQuoteBase` rejects them. `max` is unaffected: `getMaxSwappable` rewrites it into a
  `from` quote.
- **Errors arrive one at a time** as `{ error: { code, message, min?, max? } }`, not as an
  `errors` array, so Edge cannot prioritize among all applicable errors.
- **Unactivated XRP/HBAR destinations** are unconfirmed: nothing in the API states whether
  Swapter bundles account activation into the withdrawal.
- **`info.type: "fixed"` still answers HTTP 500** rather than a structured 4xx, which is why
  the fixed-rate attempt cannot be distinguished from a provider outage and the float
  fallback exists.

#### Testing

`verify-repo.sh` passes (prepare, eslint, `tsc`, mocha). The suite includes
`test/swapter.test.ts` covering the adapter endpoints, both range bounds, the max-swap
probe carve-out, the create limit codes with and without an echoed bound, the
unsupported-pair codes from both routes, the empty-ticker-map failure, and the Zcash
transparent-address selection, plus the `partnerJson` cases asserting against the
regenerated live fixture.

**Swaps have been executed end to end in the app**, on the iOS simulator against the live
API with the real plugin linked in via `updot`. The most recent run was a max
DOGE to LTC swap: the quote scene rendered `481.55 DOGE ($33.81) to 0.739 LTC ($32.91)`
labelled *Powered by Swapter*, confirmed through the slider to the success scene, and the
transaction record reads `Swap Funds / Exchange:To LTC / 481.55488751 Ð (+2.568422 fee) /
Fixed Quote`, the `Fixed Quote` label confirming the `type: 'fix'` path rather than the
float fallback. Earlier rounds executed ETH (Arbitrum) to LTC, LTC to DOGE, LTC to DGB and
DGB to LTC, so both directions of the LTC pairs and several deposit assets have been
exercised. Screenshots are attached in a PR comment.

#### Follow-ups

- `swapter` is registered but not yet enabled anywhere: the info server serves no Swapter
  init options, so the GUI needs `SWAPTER_INIT` before it can surface.
- Consider a `mapctl` synchronizer for Swapter. `/data/coins` needs no API key, so it is a
  clean fit and would catch renames automatically instead of silently dropping a chain. Not
  done here because `mapctl update-mappings` constructs every provider's synchronizer
  eagerly and aborts without all of their API keys.

---

> [!NOTE]
> **Medium Risk**
> New live swap provider that creates real orders and moves user funds via external API; complexity is high (error codes, fixed/float fallback, max-swap probe) but the design mirrors existing central plugins.
> 
> **Overview**
> **Adds Swapter** as a centralized swap provider: registers `makeSwapterPlugin` in `src/index.ts`, documents the release in `CHANGELOG.md`, and wires `SWAPTER_INIT` (`apiKey`) in test config.
> 
> The plugin (`src/swap/central/swapter.ts`) quotes via Swapter’s **Edge adapter** routes (`/adapter/edge/swap/deposit-range` and `/adapter/edge/swap/create`) instead of `/v2`, loads supported tokens from **unauthenticated** `/data/coins`, and maps Edge chains in `src/mappings/swapter.ts`. Quoting is **deposit-amount only** (`quoteFor: 'from'`); it tries **fixed** rate (`type: 'fix'`) then falls back to **float** when create rejects the fixed attempt without a pair/amount error. Deposit bounds are enforced from `deposit-range` (with inward rounding to native limits); API codes map to `SwapCurrencyError`, `SwapBelowLimitError`, and `SwapAboveLimitError`. **Max swaps** use a fee probe to the user’s refund address with `skipChecks`, then a single create order. Special cases include transparent **Zcash** addresses, `SPECIAL_MAINNET_CASES` (TON/BSV/EOS), and blocking native **zkSync** via `INVALID_TOKEN_IDS`.
> 
> **Tests:** `test/swapter.test.ts` covers adapter paths, limits, max swap, error codes, empty asset cache, and Zcash; `test/partnerJson` adds Swapter cases and a regenerated `swapterMap.json` fixture.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 60d88d2132dbcbef6f88291eb92e9f013cfebbd4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/mappings/swapter.ts
+++ b/src/mappings/swapter.ts
@@ -1,0 +1,137 @@
+/**
+ * Swapter Exchange Plugin Chain Mapping
+ *
+ * See https://docs.swapter.io/ for API documentation
+ * Currency list endpoint, which must be called UNAUTHENTICATED:
+ *
+ * curl -X GET 'https://api.swapter.io/data/coins'
+ *
+ * Sending an `X-API-KEY` header here answers 401 for a valid key, a garbage key
+ * and an empty value alike, which is why the plugin fetches this one route with
+ * `publicHeaders`. A curl carrying the key reproduces an empty ticker map, not
+ * the real list.
+ *
+ * This file maps EdgeCurrencyPluginId -> Swapter network identifier
+ *
+ * Notes:
+ * - Only networks confirmed from `/data/coins` are mapped.
+ * - Unsupported or unverified chains are set to `null`.
+ * - Swapter uses custom network identifiers for some chains:
+ *     - Avalanche C-Chain -> AVAX_C
+ *     - Binance Smart Chain -> BSC
+ *     - Ethereum PoW -> ETHEREUM POW (note the space)
+ *     - Tron -> TRX
+ *     - zkSync Era -> ZKS20
+ *     - WAX -> WAXP
+ * - Some Edge plugin IDs intentionally differ from Swapter naming.
+ * - EVM chains are mapped directly because Swapter identifies them
+ *   by network name instead of chain ID.
+ * - Swapter renames network identifiers over time (Arbitrum was `ARBITRUM`,
+ *   now `ARB`). A stale code here is not caught by any type: mainnet quotes
+ *   still reach the API and fail with Swapter's "combination does not exists"
+ *   error, while token quotes throw `SwapCurrencyError` because `/data/coins`
+ *   returns no ticker set for the unknown network. Re-verify every code against
+ *   `/data/coins` before editing this file.
+ * - A chain whose Edge `currencyCode` differs from Swapter's asset code cannot
+ *   be mapped here alone: `getChainAndTokenCodes` derives the mainnet ticker
+ *   from `currencyInfo.currencyCode`. Three chains need a `SPECIAL_MAINNET_CASES`
+ *   entry in the plugin for that reason: native TON is listed as `GRAM`, Bitcoin
+ *   SV is `BSV` in Edge and `BCHSV` on Swapter (both the network and the asset),
+ *   and the EOS network lists only Vaulta's `A` since the rebrand. Swapter lists
+ *   no `BSV` network at all, so mapping it to `BSV` would break the chain
+ *   outright.
+ * - A mapped network does NOT imply the native asset is quotable. zkSync's
+ *   `ZKS20` lists the `ZK` token but no native ETH, so the network stays mapped
+ *   (tokens quote) while the native is blocked through the plugin's
+ *   `INVALID_TOKEN_IDS`.
+ */
+
+import { EdgeCurrencyPluginId } from '../util/edgeCurrencyPluginIds'
+
+export const swapter = new Map<EdgeCurrencyPluginId, string | null>()
+
+swapter.set('abstract', null)
+swapter.set('algorand', 'ALGO')
+swapter.set('amoy', null)
+swapter.set('arbitrum', 'ARB')
+swapter.set('avalanche', 'AVAX_C')
+swapter.set('axelar', 'AXL')
+swapter.set('badcoin', null)
+swapter.set('base', 'BASE')
+swapter.set('binance', null)
+swapter.set('binancesmartchain', 'BSC')
+swapter.set('bitcoin', 'BTC')
+swapter.set('bitcoincash', 'BCH')
+swapter.set('bitcoincashtestnet', null)
+swapter.set('bitcoingold', null)
+swapter.set('bitcoingoldtestnet', null)
+swapter.set('bitcoinsv', 'BCHSV')
+swapter.set('bitcointestnet', null)
+swapter.set('bitcointestnet4', null)
+swapter.set('bobevm', null)
+swapter.set('botanix', null)
+swapter.set('calibration', null)
+swapter.set('cardano', 'ADA')
+swapter.set('cardanotestnet', null)
+swapter.set('celo', 'CELO')
+swapter.set('coreum', null)
+swapter.set('cosmoshub', 'ATOM')
+swapter.set('dash', 'DASH')
+swapter.set('digibyte', 'DGB')
+swapter.set('dogecoin', 'DOGE')
+swapter.set('eboost', null)
+swapter.set('ecash', 'XEC')
+swapter.set('eos', 'EOS')
+swapter.set('ethDev', null)
+swapter.set('ethereum', 'ETH')
+swapter.set('ethereumclassic', 'ETC')
+swapter.set('ethereumpow', 'ETHEREUM POW')
+swapter.set('fantom', null)
+swapter.set('feathercoin', null)
+swapter.set('filecoin', 'FIL')
+swapter.set('filecoinfevm', null)
+swapter.set('filecoinfevmcalibration', null)
+swapter.set('fio', null)
+swapter.set('groestlcoin', null)
+swapter.set('hedera', 'HBAR')
+swapter.set('holesky', null)
+swapter.set('hyperevm', 'HYPEREVM')
+swapter.set('liberland', null)
+swapter.set('liberlandtestnet', null)
+swapter.set('litecoin', 'LTC')
+swapter.set('mayachain', null)
+swapter.set('monad', 'MONAD')
+swapter.set('monero', 'XMR')
+swapter.set('nym', null)
+swapter.set('opbnb', null)
+swapter.set('optimism', 'OP')
+swapter.set('osmosis', 'OSMO')
+swapter.set('piratechain', null)
+swapter.set('pivx', null)
+swapter.set('polkadot', 'DOT')
+swapter.set('polygon', 'POL')
+swapter.set('pulsechain', null)
+swapter.set('qtum', 'QTUM')
+swapter.set('ravencoin', null)
+swapter.set('ripple', 'XRP')
+swapter.set('rsk', null)
+swapter.set('sepolia', null)
+swapter.set('smartcash', null)
+swapter.set('solana', 'SOL')
+swapter.set('sonic', 'SONIC')
+swapter.set('stellar', 'XLM')
+swapter.set('sui', 'SUI')
+swapter.set('suitestnet', null)
+swapter.set('telos', 'TELOS')
+swapter.set('tezos', 'XTZ')
+swapter.set('thorchainrune', 'RUNE')
+swapter.set('thorchainrunestagenet', null)
+swapter.set('ton', 'TON')
+swapter.set('tron', 'TRX')
+swapter.set('ufo', null)
+swapter.set('vertcoin', null)
+swapter.set('wax', 'WAXP')
+swapter.set('zano', null)
+swapter.set('zcash', 'ZEC')
+swapter.set('zcoin', null)
+swapter.set('zksync', 'ZKS20')
--- a/src/swap/central/swapter.ts
+++ b/src/swap/central/swapter.ts
@@ -1,0 +1,747 @@
+import { ceil, floor, gt, lt } from 'biggystring'
+import {
+  asArray,
+  asEither,
+  asJSON,
+  asMaybe,
+  asNull,
+  asObject,
+  asString
+} from 'cleaners'
+import {
+  EdgeCorePluginOptions,
+  EdgeMemo,
+  EdgeSpendInfo,
+  EdgeSwapInfo,
+  EdgeSwapPlugin,
+  EdgeSwapQuote,
+  EdgeSwapRequest,
+  SwapAboveLimitError,
+  SwapBelowLimitError,
+  SwapCurrencyError
+} from 'edge-core-js/types'
+
+import { swapter as swapterMapping } from '../../mappings/swapter'
+import {
+  ChainCodeTickerMap,
+  checkInvalidTokenIds,
+  checkWhitelistedMainnetCodes,
+  CurrencyPluginIdSwapChainCodeMap,
+  denominationToNative,
+  EdgeIdSwapIdMap,
+  ensureInFuture,
+  getChainAndTokenCodes,
+  getMaxSwappable,
+  InvalidTokenIds,
+  makeSwapPluginQuote,
+  mapToRecord,
+  nativeToDenomination,
+  SwapOrder
+} from '../../util/swapHelpers'
+import { convertRequest, getAddress, memoType } from '../../util/utils'
+import { asNumberString, EdgeSwapRequestPlugin, StringMap } from '../types'
+import { asOptionalBlank } from './changenow'
+
+const pluginId = 'swapter'
+const orderUri = 'https://swapter.io/exchange-status/'
+const apiBaseUrl = 'https://api.swapter.io'
+
+/** Swapter returns no quote expiration, so bound it locally. */
+const EXPIRATION_MS = 1000 * 60 * 30
+
+/**
+ * Swapter's zkSync network (`ZKS20`) lists the `ZK` token but NOT native ETH, so
+ * a native zkSync quote reaches the API and is rejected as a coin/network
+ * combination that does not exist. Blocking the null tokenId keeps the network
+ * mapped, which is what makes its tokens quotable.
+ */
+const INVALID_TOKEN_IDS: InvalidTokenIds = {
+  from: { zksync: [null] },
+  to: { zksync: [null] }
+}
+
+/** How long a fetched `/data/coins` ticker map is reused. */
+const ASSET_CACHE_MS = 1000 * 60 * 60 // 1 hour
+
+export const swapInfo: EdgeSwapInfo = {
+  pluginId,
+  isDex: false,
+  displayName: 'Swapter',
+  supportEmail: 'support@swapter.io'
+}
+
+const asInitOptions = asObject({ apiKey: asString })
+
+/**
+ * Chains whose default address is not the one a CEX can pay or refund. Zcash
+ * lists its unified address first, which Swapter cannot send to or refund, so
+ * the deposit or the payout would be stranded. Matches `changenow` and
+ * `letsexchange`.
+ */
+const addressTypeMap: StringMap = {
+  zcash: 'transparentAddress'
+}
+
+const asSwapterNetwork = asObject({
+  network: asString,
+  contract: asEither(asNull, asString)
+})
+
+const asSwapterAssetsResponse = asObject({
+  assets: asArray(
+    asObject({
+      currency: asString,
+      networks: asObject({
+        deposit: asArray(asSwapterNetwork),
+        withdraw: asArray(asSwapterNetwork)
+      })
+    })
+  )
+})
+
+const asSwapterDepositRangeResponse = asObject({
+  min: asNumberString,
+  max: asNumberString
+})
+
+const asSwapterErrorResponse = asJSON(
+  asObject({
+    error: asObject({
+      // Swapter sends string codes today, but its amount fields are numbers, so
+      // accept either rather than silently falling back to a generic error.
+      code: asNumberString,
+      message: asMaybe(asString),
+      // Present only on the limit rejections below, carrying the bound that was
+      // violated in the deposit asset's denomination.
+      min: asMaybe(asNumberString),
+      max: asMaybe(asNumberString)
+    })
+  })
+)
+
+/**
+ * Codes meaning Swapter does not serve this pair, which happens whenever it
+ * drops or renames an asset this mapping still lists. Those are not failures,
+ * so they become `SwapCurrencyError` and Swapter drops out of the quote race.
+ *
+ * The two endpoints do NOT share an error vocabulary: the deposit range answers
+ * a bare service code for either side of the pair, while create namespaces its
+ * own. Both spellings are listed because either call can be the one that
+ * discovers the pair is gone.
+ */
+const UNSUPPORTED_PAIR_CODES = [
+  // Deposit range: coin/network combination does not exist.
+  '1',
+  '2',
+  // Create: unknown coin, unknown deposit network, unknown withdraw network.
+  'io.swapter.controller.swap.factory:1',
+  'io.swapter.controller.swap.factory:2',
+  'io.swapter.controller.swap.factory:3'
+]
+
+/**
+ * Codes `create` returns for a deposit amount outside the pair's range. They
+ * were a single code until Swapter split them, so the only way to tell a floor
+ * from a ceiling used to be string-matching their English message.
+ */
+const BELOW_MINIMUM_CODE = 'io.swapter.controller.swap.factory:6'
+const ABOVE_MAXIMUM_CODE = 'io.swapter.controller.swap.factory:9'
+
+/**
+ * Swapter's Edge-specific adapter routes. Prefer them over the `/v2` originals
+ * for two reasons that are not cosmetic:
+ *
+ * - They serialize every numeric field as a string. The `/v2` routes return
+ *   unquoted JSON numbers, so an 18-decimal amount loses its low digits in
+ *   `JSON.parse` before this plugin can read it.
+ * - The range is authoritative. `/v2/swap/min-amount` reports a LOWER minimum
+ *   than create actually enforces (0.014153 vs 0.016006 ETH on ETH->LTC), so
+ *   quoting against it let an amount pass this plugin's own check and then fail
+ *   create.
+ */
+const DEPOSIT_RANGE_PATH = '/adapter/edge/swap/deposit-range'
+const CREATE_PATH = '/adapter/edge/swap/create'
+
+const asSwapterCreateResponse = asObject({
+  uid: asString,
+  deposit: asObject({
+    address: asString,
+    /**
+     * Memo-based chains (XRP, XLM, TON…) carry a destination tag that Swapter
+     * may return as a number. `asMaybe(asString)` would swallow that into
+     * `undefined` and build a spend with NO memo, stranding the deposit, so
+     * accept number-or-string and treat only null/blank as absent.
+     */
+    memo: asOptionalBlank(asNumberString)
+  }),
+  withdraw: asObject({
+    amount: asObject({
+      expected: asNumberString
+    })
+  })
+})
+
+/**
+ * Swapter's create endpoint spells the fixed-rate type `fix`, NOT `fixed`. An
+ * unrecognized type is not rejected as a bad request: the server answers 500
+ * (`io.swapter.controller.swap.factory:7`), which is indistinguishable from a
+ * pair it cannot fix, so sending `fixed` silently downgraded EVERY quote to the
+ * floating fallback below.
+ */
+type SwapterSwapType = 'float' | 'fix'
+
+/**
+ * Wraps the Edge error from a create call Swapter REJECTED (non-ok response, so
+ * no order was created). The fixed-then-float fallback retries ONLY on this, so
+ * a failure after a successful create never spawns a second order.
+ */
+class SwapterCreateRejected extends Error {
+  readonly edgeError: Error
+  constructor(edgeError: Error) {
+    super('Swapter create rejected')
+    this.name = 'SwapterCreateRejected'
+    this.edgeError = edgeError
+  }
+}
+
+export const MAINNET_CODE_TRANSCRIPTION: CurrencyPluginIdSwapChainCodeMap = mapToRecord(
+  swapterMapping
+)
+
+/**
+ * Overrides for mainnet assets whose Swapter coin code is NOT the Edge mainnet
+ * ticker. `getChainAndTokenCodes` derives a mainnet coin from
+ * `currencyInfo.currencyCode`, which is wrong whenever Swapter uses a different
+ * symbol for the same asset. Without the override the quote sends a coin
+ * Swapter does not list, so `deposit-range` and `create` both reject the pair
+ * with their unsupported-pair code and Swapter drops out of the quote race.
+ *
+ * - Native TON (Edge `TON`) is listed as `GRAM`.
+ * - Bitcoin SV is `BSV` in Edge and `BCHSV` on Swapter, network and asset alike.
+ * - The EOS network carries only Vaulta's `A` since the rebrand; Edge still
+ *   calls the chain's native asset `EOS`.
+ */
+const SPECIAL_MAINNET_CASES: EdgeIdSwapIdMap = new Map([
+  ['ton', new Map([[null, { chainCode: 'TON', tokenCode: 'GRAM' }]])],
+  ['bitcoinsv', new Map([[null, { chainCode: 'BCHSV', tokenCode: 'BCHSV' }]])],
+  ['eos', new Map([[null, { chainCode: 'EOS', tokenCode: 'A' }]])]
+])
+
+/**
+ * Errors that describe the PAIR or the AMOUNT rather than the swap type, so
+ * retrying the same request as a floating order earns the identical rejection.
+ */
+const PAIR_LEVEL_ERROR_NAMES = [
+  'SwapCurrencyError',
+  'SwapBelowLimitError',
+  'SwapAboveLimitError'
+]
+
+/**
+ * edge-core-js builds its swap errors with constructors that RETURN a plain
+ * `Error` carrying a `name`, never `this`, so `instanceof` against them is
+ * always false and silently classifies every one of them as a generic failure.
+ * Match the name instead, which is the field core's own `asMaybeSwap*Error`
+ * cleaners key off.
+ */
+const isPairLevelError = (error: unknown): boolean =>
+  error instanceof Error && PAIR_LEVEL_ERROR_NAMES.includes(error.name)
+
+/**
+ * Convert a deposit-side limit from Swapter's denomination to the native units
+ * Edge's limit errors carry. `denominationToNative` is a plain multiply, so a
+ * bound with more decimals than the asset's denomination lands on a fractional
+ * value, and native amounts must be integer atomic units. Round each bound
+ * INWARD — minimums up, maximums down — so the rounding can never widen the
+ * range past what Swapter accepts and let a doomed amount through.
+ */
+const toNativeLimit = (
+  request: EdgeSwapRequestPlugin,
+  amount: string,
+  bound: 'min' | 'max'
+): string => {
+  const native = denominationToNative(
+    request.fromWallet,
+    amount,
+    request.fromTokenId
+  )
+  return bound === 'min' ? ceil(native, 0) : floor(native, 0)
+}
+
+export function makeSwapterPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
+  const { io, log } = opts
+
+  // Per-plugin rather than module-level: the core builds one plugin instance, so
+  // the cache lifetime is unchanged in production, while a second instance (a
+  // test) can no longer inherit a map the first one stamped.
+  let chainCodeTickerMap: ChainCodeTickerMap = new Map()
+  let lastUpdated = 0
+  const { fetchCors = io.fetch } = io
+  const initOptions = asInitOptions(opts.initOptions)
+
+  const headers = {
+    'Content-Type': 'application/json',
+    'X-API-KEY': initOptions.apiKey,
+    Accept: 'application/json'
+  }
+
+  /**
+   * `/data/coins` is public and rejects any request that carries an
+   * `X-API-KEY` header at all — it answers 401 for a valid-format key, a
+   * garbage key and an empty value alike, but 200 with no key. Sending the key
+   * here would leave the ticker map permanently empty, which is silent: the
+   * failure is only warned, and every token quote then dies in
+   * `getChainAndTokenCodes` with `SwapCurrencyError`.
+   */
+  const publicHeaders = {
+    'Content-Type': 'application/json',
+    Accept: 'application/json'
+  }
+
+  async function fetchSupportedAssets(): Promise<void> {
+    if (lastUpdated > Date.now() - ASSET_CACHE_MS) return
+
+    try {
+      const response = await fetchCors(apiBaseUrl + '/data/coins', {
+        headers: publicHeaders
+      })
+
+      if (!response.ok) {
+        const message = await response.text()
+        throw new Error(message)
+      }
+
+      const json = await response.json()
+      const { assets } = asSwapterAssetsResponse(json)
+
+      const chaincodeArray = Object.values(MAINNET_CODE_TRANSCRIPTION)
+      const out: ChainCodeTickerMap = new Map()
+
+      for (const asset of assets) {
+        for (const depositNetwork of asset.networks.deposit) {
+          const canWithdraw = asset.networks.withdraw.some(
+            withdrawNetwork =>
+              withdrawNetwork.network === depositNetwork.network &&
+              withdrawNetwork.contract === depositNetwork.contract
+          )
+
+          if (!canWithdraw) continue
+          if (!chaincodeArray.includes(depositNetwork.network)) continue
+
+          const tokenCodes = out.get(depositNetwork.network) ?? []
+
+          tokenCodes.push({
+            tokenCode: asset.currency,
+            contractAddress: depositNetwork.contract ?? null
+          })
+
+          out.set(depositNetwork.network, tokenCodes)
+        }
+      }
+
+      // Stamp the cache ONLY on a result that carries assets. A successful but
+      // empty response (an outage answering `{ assets: [] }`, or a snapshot
+      // whose networks no longer intersect this mapping) would otherwise be
+      // cached for the full TTL, and every token quote in that hour would fail
+      // as an unsupported pair rather than retrying.
+      if (out.size > 0) {
+        chainCodeTickerMap = out
+        lastUpdated = Date.now()
+      } else {
+        log.warn('Swapter: /data/coins returned no usable assets')
+      }
+    } catch (e: unknown) {
+      log.warn('Swapter: Error updating supported assets', e)
+    }
+  }
+
+  /**
+   * Translate a non-ok Swapter response into the right Edge error. An
+   * unquotable pair becomes `SwapCurrencyError` so the GUI drops Swapter for
+   * this request; anything else is a real provider failure and surfaces as one.
+   */
+  const swapterError = (
+    stage: string,
+    status: number,
+    text: string,
+    request: EdgeSwapRequestPlugin,
+    logError: boolean = true
+  ): Error => {
+    const errorResponse = asMaybe(asSwapterErrorResponse)(text)
+    if (errorResponse != null) {
+      const { code, min, max } = errorResponse.error
+
+      if (UNSUPPORTED_PAIR_CODES.includes(code)) {
+        return new SwapCurrencyError(swapInfo, request)
+      }
+
+      // The deposit range is quoted live and moves with the rate, so an amount
+      // that cleared the range check above can still be out of bounds by the
+      // time `create` runs. Carry the bound Swapter rejected against, so the
+      // user is told the real limit instead of a generic provider failure.
+      // Map the code even when the bound is missing: core's limit errors take
+      // `undefined` and render without a number. Requiring the bound would drop
+      // a bound-less rejection to a generic error, which is not pair-level, so
+      // the fixed-then-float fallback would spend a second create retrying the
+      // same amount Swapter just refused.
+      if (code === BELOW_MINIMUM_CODE) {
+        return new SwapBelowLimitError(
+          swapInfo,
+          min != null ? toNativeLimit(request, min, 'min') : undefined
+        )
+      }
+      if (code === ABOVE_MAXIMUM_CODE) {
+        return new SwapAboveLimitError(
+          swapInfo,
+          max != null ? toNativeLimit(request, max, 'max') : undefined
+        )
+      }
+    }
+    // The fixed-rate attempt is expected to fail for pairs Swapter cannot fix
+    // (it 500s), and the caller falls back to a float quote, so suppress the
+    // warning there rather than logging a provider error on every quote.
+    if (logError) log.warn(`Swapter ${stage} API error response:`, text)
+    return new Error(`Swapter ${stage} returned error code ${status}`)
+  }
+
+  /**
+   * Shared quote setup: gate the direction, resolve Swapter's asset codes and
+   * both addresses, and enforce the pair's deposit range. Creates NO order, so
+   * both the probe and the real quote can call it.
+   *
+   * `enforceMax` is false only for the probe: a max-swap probe deliberately
+   * quotes the whole pre-fee balance to discover the ceiling, so a balance above
+   * the range must clamp through `getMaxSpendable` rather than throw
+   * `SwapAboveLimitError` on an amount the user never asked to send.
+   */
+  const fetchQuoteBase = async (
+    request: EdgeSwapRequestPlugin,
+    enforceMax: boolean
+  ): Promise<{
+    swapterCodes: {
+      fromCurrencyCode: string
+      fromMainnetCode: string
+      toCurrencyCode: string
+      toMainnetCode: string
+    }
+    sourceAmount: string
+    fromAddress: string
+    toAddress: string
+  }> => {
+    const { fromWallet, toWallet, quoteFor } = request
+
+    // Swapter's create endpoint only accepts a deposit amount, so reverse
+    // quotes are unsupported. `max` never reaches here: `getMaxSwappable`
+    // rewrites it into a `from` quote before either quote path runs.
+    if (quoteFor !== 'from') {
+      throw new SwapCurrencyError(swapInfo, request)
+    }
+
+    const swapterCodes = await getChainAndTokenCodes(
+      request,
+      swapInfo,
+      chainCodeTickerMap,
+      MAINNET_CODE_TRANSCRIPTION,
+      SPECIAL_MAINNET_CASES
+    )
+
+    // Grab addresses:
+    const [fromAddress, toAddress] = await Promise.all([
+      getAddress(fromWallet, addressTypeMap[fromWallet.currencyInfo.pluginId]),
+      getAddress(toWallet, addressTypeMap[toWallet.currencyInfo.pluginId])
+    ])
+
+    // Convert the native amount to a denomination:
+    const sourceAmount = nativeToDenomination(
+      fromWallet,
+      request.nativeAmount,
+      request.fromTokenId
+    )
+
+    const depositRangeResponse = await fetchCors(
+      apiBaseUrl + DEPOSIT_RANGE_PATH,
+      {
+        headers,
+        method: 'POST',
+        body: JSON.stringify({
+          deposit: {
+            coin: swapterCodes.fromCurrencyCode,
+            network: swapterCodes.fromMainnetCode
+          },
+          withdraw: {
+            coin: swapterCodes.toCurrencyCode,
+            network: swapterCodes.toMainnetCode
+          }
+        })
+      }
+    )
+
+    if (!depositRangeResponse.ok) {
+      const text = await depositRangeResponse.text()
+      throw swapterError(
+        'deposit range',
+        depositRangeResponse.status,
+        text,
+        request
+      )
+    }
+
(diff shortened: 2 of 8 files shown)


src/swap/central/swapter.ts
```diff
@@ -0,0 +1,432 @@
...
+    'Content-Type': 'application/json',
+    'X-API-KEY': initOptions.apiKey,
+    Accept: 'application/json'
+  }
+
+  async function fetchSupportedAssets(): Promise<void> {
+    if (lastUpdated > Date.now() - EXPIRATION) return
+
+    try {
+      const response = await fetchCors(apiBaseUrl + '/data/coins', { headers })
```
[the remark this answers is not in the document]
j0ntz: Fixed in `2110233`, and confirmed against the live API.

`/data/coins` answers 401 for *any* `X-API-KEY` value (valid-format key, garbage, even an empty value) and 200 with no key at all, so this was not a bad-credential symptom: the endpoint rejects the header outright. It now uses unauthenticated headers.

I had originally left the header on because the published docs list that endpoint as authenticated, but the failure is silent (warn-only, then an empty ticker map), and sending no key is provably correct today and can't be worse than sending one. Verified by driving the real plugin: a BTC -> USDC(ERC20) token quote previously died in `getChainAndTokenCodes`; it now gets a 200 from `/data/coins`, resolves the token, and proceeds to `create`.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,432 @@
...
+      throw error
+    }
+
+    const fromNativeAmount = request.nativeAmount
+
+    const toNativeAmount = denominationToNative(
+      toWallet,
+      quoteReply.withdraw.amount.expected,
+      request.toTokenId
+    )
```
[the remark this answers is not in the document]
j0ntz: Fixed in `2110233`. Both conversions now go through `round(..., 0)`.

Real: `denominationToNative` is a plain `mul`, so a provider amount with more decimals than the asset's denomination yields a fractional native value, e.g. `mul('0.123456789', '100000000')` = `12345678.9`.

One correction to the report: these values cannot reach `spendTargets`. The deposit target uses `request.nativeAmount` (already integer, from the GUI). The fractional values would land in `savedAction.toAsset.nativeAmount` and in the `SwapBelowLimitError` limit, which is still wrong, so the fix stands.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,432 @@
...
+      throw new Error(`Swapter returned error code ${response.status}`)
+    }
+    const responseJson = await response.json()
+
+    let quoteReply
+    try {
+      quoteReply = asSwapterCreateResponse(responseJson)
+    } catch (error) {
+      log.warn('Unexpected Swapter API response:', JSON.stringify(responseJson))
+      throw error
```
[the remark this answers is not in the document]
j0ntz: Fixed in `2110233`: now `catch (error: unknown)`, per the project's catch-typing standard.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,432 @@
...
+          network: swapterCodes.toMainnetCode,
+          address: toAddress,
+          memo: null
+        }
+      })
+    })
+
+    if (!response.ok) {
+      const text = await response.text()
+      log.warn('Swapter API error response:', text)
```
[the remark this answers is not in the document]
j0ntz: Not changing this one, though `2110233` did narrow it.

This is the repo-wide convention rather than something new here: `src/swap/central/template.ts` (the documented starting point, step 8 of `docs/CREATING_AN_EXCHANGE_PLUGIN.md`) prescribes exactly this pair: `log.warn('<X> API error response:', text)` and `log.warn('Unexpected <X> API response:', JSON.stringify(responseJson))`, and every central plugin follows it, including the most recent one (`nym` logs `\`NYM ${stage} error ${response.status}: ${text}\``). Diverging in this one plugin would be inconsistent without making the class of data any less logged elsewhere.

On the specifics: the error-text log is a provider error body, not user metadata. The `JSON.stringify(responseJson)` log fires only when the response cleaner *throws*, i.e. Swapter returned a shape we don't understand. That is precisely the diagnostic worth having, and doubly so here: this cleaner has never seen a live `create` payload (the key on this task is unauthorized, so `create` only ever returns 401), so an unexpected-response dump is the thing that will explain the first real failure. The addresses it could contain are the user's own deposit/payout addresses, already stored in their transaction metadata.

`2110233` does reduce the surface: an unquotable pair now returns `SwapCurrencyError` before any logging, so the error-text path only fires for genuinely unexpected failures.

Flagging for a reviewer to overrule. If scrubbing wins, it should be a repo-wide change to `template.ts` and the other plugins rather than this PR alone.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,432 @@
...
+        })
+      }
+    )
+
+    if (!minAmountResponse.ok) {
+      const text = await minAmountResponse.text()
+      log.warn('Swapter min amount API error response:', text)
+      throw new Error(
+        `Swapter min amount returned error code ${minAmountResponse.status}`
+      )
```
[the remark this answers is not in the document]
j0ntz: Fixed in `2110233`. This was a real one.

Non-ok responses from `min-amount` and `create` now go through a shared handler that cleans Swapter's error body and maps code `1` (deposit side) / `2` (withdraw side), its 'combination does not exists' codes, to `SwapCurrencyError`, so the GUI simply omits Swapter for that pair. Anything else still surfaces as a real provider failure.

This matters more than it looks: Swapter renames networks over time (this PR fixes nine mappings that went stale exactly that way), so a mapped-but-unquotable pair is the expected steady state, not an edge case. Verified live by pinning `polygon` back to the old `POLYGON` code and driving the real plugin:

```
before: Error: Swapter min amount returned error code 400
after:  SwapCurrencyError: Swapter does not support bitcoin:null to polygon:null
```

src/swap/central/swapter.ts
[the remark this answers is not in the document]
j0ntz: Already fixed in `2110233` (same issue Bugbot raised above): `/data/coins` now uses unauthenticated headers.

Your analysis of the failure mode matches what I measured: native swaps still work because `getChainAndTokenCodes` synthesizes the mainnet asset from the mapping alone, and only token quotes that depend on the fetched contracts break. Confirmed by driving a BTC -> USDC(ERC20) quote through the real plugin before and after.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,488 @@
...
+ * this mapping still lists.
+ */
+const UNSUPPORTED_PAIR_CODES = ['1', '2']
+
+const asSwapterCreateResponse = asObject({
+  uid: asString,
+  deposit: asObject({
+    address: asString,
+    memo: asMaybe(asString)
+  }),
```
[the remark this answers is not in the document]
j0ntz: Fixed in `5235c38`. This one was real and the fund-loss reasoning is right.

`deposit.memo` is now `asOptionalBlank(asNumberString)`, reusing the existing helper from `changenow.ts` (`letsexchange` imports it the same way). Verified the cleaner directly:

| `deposit.memo` | before: `asMaybe(asString)` | after: `asOptionalBlank(asNumberString)` |
|---|---|---|
| `null` | `undefined` | `undefined` |
| `""` | `""` | `undefined` |
| `"abc123"` | `"abc123"` | `"abc123"` |
| `1234567890` | **`undefined`** | `"1234567890"` |
| `0` | **`undefined`** | `"0"` |

A numeric destination tag was silently becoming no memo at all. Destination tag `0` is also a valid XRP tag and was being dropped the same way.

This also picks up the empty-string case from the sibling thread, and it removes the silent swallow: an unparseable memo now throws through the response cleaner and gets logged rather than quietly producing a memo-less spend.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,452 @@
...
+    )
+    const memos: EdgeMemo[] =
+      quoteReply.deposit.memo == null
+        ? []
+        : [
+            {
+              type: memoType(fromWallet.currencyInfo.pluginId),
+              value: quoteReply.deposit.memo
+            }
+          ]
```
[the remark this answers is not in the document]
j0ntz: Fixed in `5235c38`, covered by the same change as the numeric-memo thread.

`asOptionalBlank(asNumberString)` treats `null` and `""` alike as absent, which is exactly the sibling behavior you're pointing at (`changenow` uses `asOptionalBlank` for `payinExtraId`). Confirmed: `""` now cleans to `undefined` instead of becoming an empty `EdgeMemo`.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,488 @@
...
+})
+
+const asSwapterErrorResponse = asJSON(
+  asObject({
+    error: asObject({
+      code: asString,
+      message: asMaybe(asString)
+    })
+  })
+)
```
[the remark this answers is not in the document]
j0ntz: Fixed in `5235c38`: `error.code` is now `asNumberString`, so numeric or string codes both classify. `UNSUPPORTED_PAIR_CODES` stays `['1', '2']` since `asNumberString` normalizes to a string.

For what it's worth, the live `min-amount` 400 does return a string today:
`{"error":{"code":"1","service":"provider.rates","message":"Deposit coin and network combination does not exists."}}`
But the `create` error shape is unverified (that endpoint only returns 401 with the key on this task), so accepting either is the right call rather than assuming it matches.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,501 @@
...
+    }
+
+    const minAmountJson = await minAmountResponse.json()
+    const { amount: minAmount } = asSwapterMinAmountResponse(minAmountJson)
+
+    if (lt(sourceAmount, minAmount)) {
+      // `denominationToNative` is a plain multiply, so a provider amount with
+      // more decimals than the asset's denomination yields a fractional native
+      // value. Native amounts must be integer atomic units.
+      const minNativeAmount = round(
```
peachbits: should use ceil instead of round

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,501 @@
...
+
+  const out: EdgeSwapPlugin = {
+    swapInfo,
+
+    async fetchSwapQuote(
+      req: EdgeSwapRequest,
+      userSettings: Object | undefined
+    ): Promise<EdgeSwapQuote> {
+      const request = convertRequest(req)
+      const settings = asUserSettings(userSettings ?? {})
```
peachbits: I don't think the settings are used? convention is to return a fixed rate if found, otherwise fall back and return a floating quote. If the gui doesn't pass any settings the plugin will be stuck always returning floating rates

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,501 @@
...
+    try {
+      quoteReply = asSwapterCreateResponse(responseJson)
+    } catch (error: unknown) {
+      log.warn('Unexpected Swapter API response:', JSON.stringify(responseJson))
+      throw error
+    }
+
+    const fromNativeAmount = request.nativeAmount
+
+    const toNativeAmount = round(
```
peachbits: use floor instead

peachbits (changes requested):

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,501 @@
...
+
+    let quoteReply
+    try {
+      quoteReply = asSwapterCreateResponse(responseJson)
+    } catch (error: unknown) {
+      log.warn('Unexpected Swapter API response:', JSON.stringify(responseJson))
+      throw error
+    }
+
+    const fromNativeAmount = request.nativeAmount
```
peachbits, line 562: **Suggestion:** consider asserting that the deposit amount Swapter echoes back matches what we're about to send.

The plugin only enforces a minimum — there's no maximum check and no `SwapAboveLimitError`. If Swapter clamps an over-max deposit to its ceiling instead of returning an error, the order gets written for the clamped amount while the spend target still sends `request.nativeAmount`, and the excess is refunded or credited at whatever rate Swapter picks. `nym.ts:266-268` guards against exactly this silent-clamp behavior.

If `/v2/swap/create` echoes a deposit amount (the withdraw side has `withdraw.amount.expected`, so a symmetric `deposit.amount.expected` seems likely), adding it to `asSwapterCreateResponse` and comparing against `sourceAmount` with `eq` would catch it. `eq` is decimal-aware, so a differing scale (`1.20` vs `1.2`) won't false-trip.

One thing to verify before shipping the check: `sourceAmount` carries up to 19 decimal places for an 18-decimal token (`div(nativeAmount, multiplier, multiplier.length)`). If Swapter float64-parses that, the echo comes back a few wei off and a strict `eq` would reject a perfectly good order. Worth confirming against a real 18-decimal-token order — and if the echo does lose precision, trimming `sourceAmount` to the asset's actual decimal count before sending fixes both the mismatch and the check.

src/swap/central/swapter.ts
j0ntz: Fixed: `minNativeAmount` now uses `ceil` instead of `round`, so the enforced minimum never rounds below Swapter's floor.

src/swap/central/swapter.ts
j0ntz: Fixed: `toNativeAmount` now uses `floor` instead of `round`, so the displayed receive amount is never rounded up above what Swapter actually sends.

src/swap/central/swapter.ts
j0ntz: Fixed. Dropped the `userSettings.swapType` selection (the GUI never passes it, so the plugin was stuck on float) and now try a fixed-rate order first, falling back to a floating quote, matching the changenow convention. Swapter's create endpoint 500s on `type: 'fixed'` for pairs it cannot fix, so that attempt's create error is suppressed and the float fallback is used. Verified in-app: an ETH(Arbitrum) to LTC swap executed to the success scene through the float fallback.

src/swap/central/swapter.ts
j0ntz: Checked this against the live API before adding it, and it would do more harm than good here. Swapter float64-parses the deposit amount, so the echo does not round-trip for high-precision tokens: sending `0.123456789012345678` echoes back `0.12345678901234568`, and `0.015500000000000001` echoes `0.015500000000000002`. A strict `eq(echoedDeposit, sourceAmount)` would false-reject any 18-decimal-token order, and trimming `sourceAmount` to the asset's 18 decimals does not fix it since float64 cannot hold 18 significant digits. Swapter also serves only float orders for the pairs tested (`type: 'fixed'` returns HTTP 500), and a float deposit is not rate-locked, so a wei-level mismatch does not lock a bad rate. An over-max probe (1,000,000 ETH) echoed the full amount un-clamped, so there is no enforced maximum to guard with a `SwapAboveLimitError`. Leaving the assertion out; worth revisiting if Swapter adds real fixed-rate support and an enforced max.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,521 @@
...
+      // the rate from `userSettings` instead left the plugin stuck on float,
+      // since the GUI never passes a `swapType`. Swapter's create endpoint 500s
+      // on `type: 'fixed'` for pairs it cannot fix, so that attempt is expected
+      // to fail for those and its create error is not logged.
+      let swapOrder: SwapOrder
+      try {
+        swapOrder = await fetchSwapQuoteInner(newRequest, 'fixed', false)
+      } catch (error) {
+        swapOrder = await fetchSwapQuoteInner(newRequest, 'float')
+      }
```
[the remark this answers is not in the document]
j0ntz: Fixed. The fixed-rate create now throws a `SwapterCreateRejected` sentinel only when Swapter rejects the create outright (non-ok response, no order created); the fallback retries as float ONLY on that sentinel. A failure after a successful create (e.g. an unparseable response) is thrown bare and never spawns a second order.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,521 @@
...
+      // Prefer a fixed-rate order, falling back to a floating quote, the
+      // convention the other central plugins follow (see changenow). Selecting
+      // the rate from `userSettings` instead left the plugin stuck on float,
+      // since the GUI never passes a `swapType`. Swapter's create endpoint 500s
+      // on `type: 'fixed'` for pairs it cannot fix, so that attempt is expected
+      // to fail for those and its create error is not logged.
+      let swapOrder: SwapOrder
+      try {
+        swapOrder = await fetchSwapQuoteInner(newRequest, 'fixed', false)
+      } catch (error) {
```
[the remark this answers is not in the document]
j0ntz: Fixed: the fallback now uses `catch (error: unknown)` (and the inner float catch uses an optional binding), per the repo's catch-typing convention.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,521 @@
...
+        {
+          nativeAmount: request.nativeAmount,
+          publicAddress: fromAddress
+        }
+      ],
+      networkFeeOption: 'high',
+      assetAction: {
+        assetActionType: 'swap'
+      }
+    }
```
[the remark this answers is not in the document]
j0ntz: Fixed: the max-fee probe's spendInfo now sets `skipChecks: true`, matching the NYM fix, so EVM max swaps no longer fail with SpendToSelfError. Verified in-app: a MAX swap from an Arbitrum wallet now reaches a real Swapter quote (below-limit at Swapter's 26.23 USDC floor) instead of erroring "Spending to self".

src/swap/central/swapter.ts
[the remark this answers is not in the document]
j0ntz: Fixed: when both the fixed and float creates fail, the fixed-rate error is now rethrown (`error.edgeError`), so limit/currency errors keep their ranking in `pickBestError`, matching the sibling plugins.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,552 @@
...
+        // a below-limit/unsupported error from the shared setup) propagates
+        // unchanged, so it is never masked and no second order is created.
+        if (!(error instanceof SwapterCreateRejected)) throw error
+        try {
+          swapOrder = await fetchSwapQuoteInner(newRequest, 'float')
+        } catch {
+          // Rethrow the fixed-rate error so limit/currency errors keep their
+          // ranking in `pickBestError`, matching the sibling plugins.
+          throw error.edgeError
+        }
```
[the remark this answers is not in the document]
j0ntz: Fixed. The fallback now surfaces the FLOAT error whenever the float attempt fails after creating an order (or any non-rejection failure), so a live float order is never masked behind the fixed-rate rejection. The fixed-rate error is rethrown only when the float create was ALSO cleanly rejected (no order exists), which preserves limit/currency error ranking. The catch binding is now typed `floatError: unknown`.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,557 @@
...
+    if (quoteFor !== 'from') {
+      throw new SwapCurrencyError(swapInfo, request)
+    }
+
+    const swapterCodes = await getChainAndTokenCodes(
+      request,
+      swapInfo,
+      chainCodeTickerMap,
+      MAINNET_CODE_TRANSCRIPTION
+    )
```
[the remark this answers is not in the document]
j0ntz: Fixed: added a `SPECIAL_MAINNET_CASES` override pinning native TON to Swapter's `GRAM` coin (verified live: native TON is listed as `GRAM` with a null contract), passed into `getChainAndTokenCodes`. Audited all 38 other mapped networks against Swapter's native coin codes and TON was the only mismatch (WAX's `WAXP` matches Edge's mainnet code). partnerJson tests pass.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,557 @@
...
+      try {
+        swapOrder = await fetchSwapQuoteInner(newRequest, 'fixed', false)
+      } catch (error: unknown) {
+        // Only fall back to a floating quote when Swapter REJECTED the fixed
+        // create (no order was made). Any other error (an order was created, or
+        // a below-limit/unsupported error from the shared setup) propagates
+        // unchanged, so it is never masked and no second order is created.
+        if (!(error instanceof SwapterCreateRejected)) throw error
+        try {
+          swapOrder = await fetchSwapQuoteInner(newRequest, 'float')
```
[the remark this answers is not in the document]
j0ntz: Fixed: an unsupported-pair error (`SwapCurrencyError`, codes 1/2) is now thrown directly instead of being wrapped in `SwapterCreateRejected`, so it propagates with no wasteful float retry. Only a fixed-specific rejection (Swapter's `type: 'fixed'` 500) is wrapped for the fallback.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,557 @@
...
+      UNSUPPORTED_PAIR_CODES.includes(errorResponse.error.code)
+    ) {
+      return new SwapCurrencyError(swapInfo, request)
+    }
+    // The fixed-rate attempt is expected to fail for pairs Swapter cannot fix
+    // (it 500s), and the caller falls back to a float quote, so suppress the
+    // warning there rather than logging a provider error on every quote.
+    if (logError) log.warn(`Swapter ${stage} API error response:`, text)
+    return new Error(`Swapter ${stage} returned error code ${status}`)
+  }
```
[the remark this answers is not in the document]
j0ntz: Below-minimum is already surfaced as `SwapBelowLimitError`: `fetchQuoteBase` fetches the pair minimum and throws it client-side before any create, so the common limit case keeps its ranking. Over-max does not occur: Swapter enforces no maximum (a 1,000,000-unit deposit echoes back un-clamped, confirmed live), so there is no `SwapAboveLimitError` to raise. The residual is a rare create-side race (the min shifting between the min-amount call and create), where a generic error is acceptable.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,557 @@
...
+    }
+    const responseJson = await response.json()
+
+    let quoteReply
+    try {
+      quoteReply = asSwapterCreateResponse(responseJson)
+    } catch (error: unknown) {
+      log.warn('Unexpected Swapter API response:', JSON.stringify(responseJson))
+      throw error
+    }
```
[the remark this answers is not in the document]
j0ntz: Inherent to Swapter's design and pre-existing: `/v2/swap/create` IS the quote (no separate quote endpoint) and there is no cancel endpoint, so every quote necessarily creates an order and any create-then-fail leaves one that expires unfunded. Wrapping the parse failure in `SwapterCreateRejected` would be worse, re-introducing the double-order the fallback was just fixed to avoid. The unparseable-response order is never funded (the spend is never built) and expires on Swapter's side.

j0ntz:
## 📸 Test evidence (review-round re-verification, HEAD 441a45e)

**swapter quote**

swapter quote

**swap success**

swap success

**txn details**

txn details

**max skipchecks**

max skipchecks

j0ntz:
## 📸 Test evidence

**swapter quote**

swapter quote

**swap success**

swap success

**swap tx recorded**

swap tx recorded

src/mappings/swapter.ts
```diff
@@ -0,0 +1,127 @@
...
+swapter.set('ripple', 'XRP')
+swapter.set('rsk', null)
+swapter.set('sepolia', null)
+swapter.set('smartcash', null)
+swapter.set('solana', 'SOL')
+swapter.set('sonic', 'SONIC')
+swapter.set('stellar', 'XLM')
+swapter.set('sui', 'SUI')
+swapter.set('suitestnet', null)
+swapter.set('telos', 'TELOS')
```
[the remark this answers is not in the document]
j0ntz: Rejected: the premise is backwards. Swapter lists TWO separate Telos networks, and `TELOS` is the EOSIO one.

From live `/data/coins` today, the `TLOS` asset carries both:

```
{"network":"TELOS","memo":false,"contract":"eosio.token"}
{"network":"TELOS(EVM)","memo":false,"contract":null}
```

`eosio.token` is the native EOSIO token contract. The EVM chain is the separately-named `TELOS(EVM)`, which this mapping does not use.

Confirmed by creating a real order on `TELOS` and reading back the deposit target:

```
POST /v2/swap/create  {"deposit":{"coin":"TLOS","network":"TELOS","amount":"1500"}, ...}
-> {"coin":"TLOS","network":"TELOS","address":"kcdeposit4in","memo":"2097028190"}
```

`kcdeposit4in` is a 12-character EOSIO account name with a numeric memo, not an `0x` address, so this is the antelope/EOSIO chain that Edge's `telos` plugin targets.

The memo is handled: `asSwapterCreateResponse` cleans `deposit.memo` and the quote attaches it as an `EdgeMemo` whose type comes from `memoType('telos')` = `text`, which matches Telos's `memoOptions` (`type: 'text'`, maxLength 256) in edge-currency-accountbased.

src/mappings/swapter.ts
```diff
@@ -0,0 +1,127 @@
...
+swapter.set('badcoin', null)
+swapter.set('base', 'BASE')
+swapter.set('binance', null)
+swapter.set('binancesmartchain', 'BSC')
+swapter.set('bitcoin', 'BTC')
+swapter.set('bitcoincash', 'BCH')
+swapter.set('bitcoincashtestnet', null)
+swapter.set('bitcoingold', null)
+swapter.set('bitcoingoldtestnet', null)
+swapter.set('bitcoinsv', 'BCHSV')
```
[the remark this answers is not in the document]
j0ntz: Rejected: `BCHSV` is currently listed, for both deposit and withdraw.

Checked against live `/data/coins` at the time of this reply (912 assets, 233 networks):

```
BCHSV   deposit=true  withdraw=true
```

Both quote directions resolve:

```
POST /v2/swap/min-amount  BCHSV/BCHSV -> LTC   200 {"amount":"1.71933614"}
POST /v2/swap/min-amount  LTC -> BCHSV/BCHSV   200 {"amount":"0.49324344"}
```

And a real create returns a valid BSV address:

```
POST /v2/swap/create {"deposit":{"coin":"BCHSV","network":"BCHSV","amount":"2"}, ...}
-> {"coin":"BCHSV","network":"BCHSV","address":"1Fgxiin7Q9A1rRnThVwZVLF4YRZLwLBj1c","memo":null}
```

What does not exist is a network literally named `BSV` (`min-amount` returns `code 1, deposit coin and network combination does not exists`), which is why the mapping goes through `SPECIAL_MAINNET_CASES`: Edge's `currencyCode` is `BSV` while Swapter names both the network and the asset `BCHSV`.

j0ntz:
## 📸 Test evidence: LTC to DGB swap on the deposit-range build (2026-08-07)

**swapter quote**

swapter quote

**swap success**

swap success

**swap tx recorded**

swap tx recorded

j0ntz:
## 📸 Test evidence: DGB to LTC swap on the adapter-create build (2026-08-11)

**s8 01 swapter quote**

s8 01 swapter quote

**s8 02 swap success**

s8 02 swap success

**s8 03 swap tx recorded**

s8 03 swap tx recorded

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,699 @@
...
+        method: 'POST',
+        body: JSON.stringify({
+          deposit: {
+            coin: swapterCodes.fromCurrencyCode,
+            network: swapterCodes.fromMainnetCode
+          },
+          withdraw: {
+            coin: swapterCodes.toCurrencyCode,
+            network: swapterCodes.toMainnetCode
+          }
```
paullinator, line 474: **Warning — #1** (API requirements)

Quote and create identify assets by ticker + Swapter network name (`deposit.coin`/`network`, `withdraw.coin`/`network`), not a unique chain identifier plus token identifier (contract). `/data/coins` exposes `contract`, but that is only used to build a local Edge tokenId→ticker map; the quote and create bodies never send the contract. `API_REQUIREMENTS.md` §1 says a separate list-assets endpoint is not sufficient. Native assets are still guessed from `currencyInfo.currencyCode` (hence `SPECIAL_MAINNET_CASES` for TON=GRAM and bitcoinsv=BCHSV). Contrast `template.ts` (`fromContractAddress`/`toContractAddress`) and `nym.ts` (`tokenId` = contract).

**Recommendation:** Change the Edge adapter (`/adapter/edge/swap/deposit-range` and `/create`) to accept chain + tokenId/contract on quote and create, and send those from `getContractAddresses(request)` (null for natives). Stop sending ticker symbols as the asset identity.

src/mappings/swapter.ts
```diff
@@ -0,0 +1,127 @@
...
+ * - Unsupported or unverified chains are set to `null`.
+ * - Swapter uses custom network identifiers for some chains:
+ *     - Avalanche C-Chain -> AVAX_C
+ *     - Binance Smart Chain -> BSC
+ *     - Ethereum PoW -> ETHEREUM POW (note the space)
+ *     - Tron -> TRX
+ *     - zkSync Era -> ZKS20
+ *     - WAX -> WAXP
+ * - Some Edge plugin IDs intentionally differ from Swapter naming.
+ * - EVM chains are mapped directly because Swapter identifies them
```
paullinator, line 27: **Warning — #2** (API requirements, EVM)

The API as used does not accept numeric EVM `chainId`, and the plugin never reads `wallet.currencyInfo.evmChainId`. EVM chains are hardcoded Edge pluginId → Swapter network name (`ethereum`→`ETH`, `arbitrum`→`ARB`, `polygon`→`POL`, …). This file states: "EVM chains are mapped directly because Swapter identifies them by network name instead of chain ID." Adding a new EVM chain (or surviving a Swapter rename) requires a plugin change. This already bit the PR (`ARBITRUM`→`ARB`, `POLYGON`→`POL`, plus later HyperEVM/Monad adds). `template.ts` returns `evmChain` whenever `evmChainId` is set and sends `fromEvmChainId`/`toEvmChainId`; `nym.ts` sends `chainNetwork` + `chainId` from `evmChainId`.

**Recommendation:** The adapter must accept the numeric EVM chainId as a way to quote. The plugin must pass `fromWallet.currencyInfo.evmChainId` / `toWallet.currencyInfo.evmChainId` and must not map Edge plugin IDs (or Swapter network names) to EVM chains. Drop EVM entries from `src/mappings/swapter.ts` the way `src/mappings/template.ts` does, so a new EVM chain needs no plugin change.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,699 @@
...
+    sourceAmount: string
+    fromAddress: string
+    toAddress: string
+  }> => {
+    const { fromWallet, toWallet, quoteFor } = request
+
+    // Swapter's create endpoint only accepts a deposit amount, so reverse
+    // quotes are unsupported. `max` never reaches here: `getMaxSwappable`
+    // rewrites it into a `from` quote before either quote path runs.
+    if (quoteFor !== 'from') {
```
paullinator, line 436: **Must fix — #3** (API requirements)

`API_REQUIREMENTS.md` §4 requires bi-directional quoting. Swapter create only accepts a deposit (`from`) amount; the plugin throws `SwapCurrencyError` for `quoteFor !== 'from'`. `max` is rewritten to `from` by `getMaxSwappable`, so exact-receive (`to`) quotes are unsupported. `template.ts` and `nym.ts` send `sourceAmount` or `destinationAmount`.

**Recommendation:** If the Edge adapter can accept a withdraw/destination amount, add a `to` path like `template.ts` / `nym.ts`. If the API is deposit-amount-only, document that reverse quotes are unsupported and get product sign-off — but the API requirements doc treats bi-directional quoting as mandatory.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,699 @@
...
+      swapInfo,
+      chainCodeTickerMap,
+      MAINNET_CODE_TRANSCRIPTION,
+      SPECIAL_MAINNET_CASES
+    )
+
+    // Grab addresses:
+    const [fromAddress, toAddress] = await Promise.all([
+      getAddress(fromWallet),
+      getAddress(toWallet)
```
paullinator: **Must fix — #4**

ZEC is mapped, but `getAddress` is called with no address type. Default `getAddress` returns `allAddresses[0]`; ZcashEngine lists `unifiedAddress` first, then sapling, then transparent. Sibling central plugins (`changenow`, `letsexchange`, `changelly`, `nexchange`, `exolix`, `xgram`) pass `{ zcash: 'transparentAddress' }`. A CEX that cannot pay or refund a unified address can strand the deposit or refund.

**Recommendation:** Add an `addressTypeMap` of `{ zcash: 'transparentAddress' }` and pass it into `getAddress` for both wallets, matching `changenow`/`letsexchange`.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,699 @@
...
+    })
+  )
+})
+
+const asSwapterDepositRangeResponse = asObject({
+  min: asNumberString,
+  max: asNumberString
+})
+
+const asSwapterErrorResponse = asJSON(
```
paullinator, line 107: **Warning — #5** (API requirements)

`API_REQUIREMENTS.md` §3 expects errors as an `errors` array of all applicable structured objects. Swapter returns a single `{ error: { code, message, min?, max? } }`. There is no `RegionRestricted` type, asset-not-supported does not name which asset, and limit errors carry only the deposit-side min/max (not both `sourceAmountLimit` and `destinationAmountLimit`). Codes are Swapter-specific rather than `OverLimitError`/`UnderLimitError`. The plugin maps what it gets; the provider API as used cannot return all errors at once for Edge to prioritize.

**Recommendation:** Adapter error bodies should match the spec: `{ errors: [ { error, sourceAmountLimit, destinationAmountLimit, ... } ] }` with all applicable errors in one response.

src/mappings/swapter.ts
```diff
@@ -0,0 +1,127 @@
...
+swapter.set('opbnb', null)
+swapter.set('optimism', 'OP')
+swapter.set('osmosis', 'OSMO')
+swapter.set('piratechain', null)
+swapter.set('pivx', null)
+swapter.set('polkadot', 'DOT')
+swapter.set('polygon', 'POL')
+swapter.set('pulsechain', null)
+swapter.set('qtum', 'QTUM')
+swapter.set('ravencoin', 'RVN')
```
paullinator: **Must fix — #8**

Several mapped chains will fail native quotes against live `/data/coins`:
- `eos`→`EOS` has no native EOS ticker (only Vaulta `A`)
- `zksync`→`ZKS20` lists only the ZK token, not native ETH
- `ravencoin`→`RVN` is absent from live coins

The mapping comments claim only confirmed networks are mapped. Native TON/BSV already needed `SPECIAL_MAINNET_CASES` for the same class of ticker mismatch.

**Recommendation:** Add `SPECIAL_MAINNET_CASES` for EOS→`A` if Vaulta is intended, or null those plugin IDs. Unmap or block native zkSync ETH. Set `ravencoin` to `null` until `/data/coins` lists it again, and regenerate `swapterMap.json` from a current snapshot.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,699 @@
...
+   * here would leave the ticker map permanently empty, which is silent: the
+   * failure is only warned, and every token quote then dies in
+   * `getChainAndTokenCodes` with `SwapCurrencyError`.
+   */
+  const publicHeaders = {
+    'Content-Type': 'application/json',
+    Accept: 'application/json'
+  }
+
+  async function fetchSupportedAssets(): Promise<void> {
```
paullinator, line 301: **Must fix — #9**

`fetchSupportedAssets` swallows every failure and also caches a successful empty map (`{ assets: [] }` or every row filtered out) for the 1-hour TTL. `getChainAndTokenCodes` still guesses native tickers from `currencyCode`, so mainnet quotes proceed, but every token quote throws `SwapCurrencyError`. The GUI treats that as an unsupported pair rather than a provider outage. Tests stub `{ assets: [] }`, which stamps the shared module cache after the first case.

**Recommendation:** If the ticker map is still empty after a fetch, fail the quote with a generic provider error instead of continuing into `getChainAndTokenCodes`. Do not set `lastUpdated` on an empty success. Keep retrying on the next quote. Reset the module cache between tests.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,699 @@
...
+
+      if (UNSUPPORTED_PAIR_CODES.includes(code)) {
+        return new SwapCurrencyError(swapInfo, request)
+      }
+
+      // The deposit range is quoted live and moves with the rate, so an amount
+      // that cleared the range check above can still be out of bounds by the
+      // time `create` runs. Carry the bound Swapter rejected against, so the
+      // user is told the real limit instead of a generic provider failure.
+      if (code === BELOW_MINIMUM_CODE && min != null) {
```
paullinator: **Must fix — #10**

Create limit codes `factory:6` and `factory:9` only become `SwapBelowLimitError`/`SwapAboveLimitError` when `min`/`max` are present. If Swapter omits the bound, `swapterError` returns a generic `Error`, `isPairLevelError` is false, and the caller wraps it in `SwapterCreateRejected`, so the float fallback retries the same doomed amount.

**Recommendation:** Map those codes regardless of bound presence: `new SwapBelowLimitError(swapInfo, min != null ? toNativeLimit(request, min, 'min') : undefined)` (and the same for max).

src/mappings/swapter.ts
```diff
@@ -0,0 +1,127 @@
+/**
+ * Swapter Exchange Plugin Chain Mapping
+ *
+ * See https://docs.swapter.io/ for API documentation
+ * Currency list endpoint:
+ *
+ * curl -X GET 'https://api.swapter.io/data/coins' \
+ *   -H 'X-API-KEY: <your-api-key>'
```
paullinator: **Must fix — #11**

The mapping header still documents fetching `/data/coins` with `X-API-KEY`, but `fetchSupportedAssets` treats that header as a hard 401. Anyone following this curl will reproduce the empty ticker-map failure the plugin already works around with `publicHeaders`.

**Recommendation:** Drop the `X-API-KEY` header from the example and note the endpoint must be called unauthenticated, matching `publicHeaders`.

src/mappings/swapter.ts
```diff
@@ -0,0 +1,127 @@
...
+swapter.set('optimism', 'OP')
+swapter.set('osmosis', 'OSMO')
+swapter.set('piratechain', null)
+swapter.set('pivx', null)
+swapter.set('polkadot', 'DOT')
+swapter.set('polygon', 'POL')
+swapter.set('pulsechain', null)
+swapter.set('qtum', 'QTUM')
+swapter.set('ravencoin', 'RVN')
+swapter.set('ripple', 'XRP')
```
paullinator, line 116: **Confirm and verify — #16** (API requirements)

`ripple` and `hedera` are mapped, but nothing in this PR shows Swapter detecting unactivated XRP/HBAR destination addresses and bundling account activation into the withdrawal. `API_REQUIREMENTS.md` §7 requires this with no extra action from Edge or the user.

**Recommendation:** Confirm with Swapter (and document in the PR) that unactivated XRP/HBAR payouts include the activation payment with no extra Edge/user step.

src/mappings/swapter.ts
```diff
@@ -0,0 +1,127 @@
...
+import { EdgeCurrencyPluginId } from '../util/edgeCurrencyPluginIds'
+
+export const swapter = new Map<EdgeCurrencyPluginId, string | null>()
+
+swapter.set('abstract', null)
+swapter.set('algorand', 'ALGO')
+swapter.set('amoy', null)
+swapter.set('arbitrum', 'ARB')
+swapter.set('avalanche', 'AVAX_C')
+swapter.set('axelar', null)
```
paullinator: **Must fix — #17**

`axelar` is `null` with no reason, but live `/data/coins` has native `AXL` on network `AXL`, matching Edge `axelar` `currencyCode` `AXL`. Other plugins (`changenow`, `rango`, `letsexchange`) map this chain.

**Recommendation:** Map `axelar` → `AXL` unless Axelar was intentionally withheld.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,699 @@
...
+      expirationDate: ensureInFuture(new Date(Date.now() + EXPIRATION_MS))
+    }
+  }
+
+  const out: EdgeSwapPlugin = {
+    swapInfo,
+
+    async fetchSwapQuote(
+      req: EdgeSwapRequest,
+      userSettings: Object | undefined
```
paullinator: **Must fix — #18**

`fetchSwapQuote` still declares unused `userSettings` after the fixed-then-float change stopped reading it. `nym`, `swapuz`, `changehero`, and `sideshift` omit it.

**Recommendation:** Remove the unused `userSettings` parameter so the signature is `fetchSwapQuote(req)` like those plugins.

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,699 @@
...
+const EXPIRATION_MS = 1000 * 60 * 30
+
+const INVALID_TOKEN_IDS: InvalidTokenIds = {
+  from: {},
+  to: {}
+}
+
+let chainCodeTickerMap: ChainCodeTickerMap = new Map()
+let lastUpdated = 0
+const EXPIRATION = 1000 * 60 * 60 // 1 hour
```
paullinator: **Must fix — #19**

`EXPIRATION_MS` (30-minute quote bound) and `EXPIRATION` (1-hour ticker-map TTL) sit a few lines apart with near-identical names. Easy to swap when touching quote expiry or the asset cache.

**Recommendation:** Rename the cache TTL to `ASSET_CACHE_MS` (keep `EXPIRATION_MS` for quotes).

src/swap/central/swapter.ts
```diff
@@ -0,0 +1,699 @@
...
+export const MAINNET_CODE_TRANSCRIPTION: CurrencyPluginIdSwapChainCodeMap = mapToRecord(
+  swapterMapping
+)
+
+/**
+ * Overrides for mainnet assets whose Swapter coin code is NOT the Edge mainnet
+ * ticker. `getChainAndTokenCodes` derives a mainnet coin from
+ * `currencyInfo.currencyCode`, which is wrong when Swapter uses a different
+ * symbol: native TON (Edge `TON`) is listed on Swapter as `GRAM`, so quotes
+ * would send an unknown coin and fail min-amount/create.
```
paullinator: **Must fix — #20**

`SPECIAL_MAINNET_CASES` still says a missing override would fail `min-amount`/create. The plugin no longer calls `/v2/swap/min-amount`; quotes go through `/adapter/edge/swap/deposit-range` and create.

**Recommendation:** Refer to deposit-range/create, not min-amount. Also mention TON→GRAM alongside BSV→BCHSV in `src/mappings/swapter.ts`.

paullinator (changes requested):
Code review for Swapter integration (issues #1, #2, #3, #4, #5, #8, #9, #10, #11, #16, #17, #18, #19, #20).

See inline comments for disposition (Warning / Must fix / Confirm and verify) and full details.

src/swap/central/swapter.ts
j0ntz: Not fixable on the client: `deposit-range` and `create` accept only `coin` + `network`, with no contract or chain-identifier field to send, so the plugin has nothing to pass even though it already resolves the contract locally. This needs an adapter change on Swapter's side, so it stays as-is in this PR.

src/mappings/swapter.ts
j0ntz: Same adapter limitation. The routes key on Swapter's own network names and reject a numeric chainId, so passing `evmChainId` has nowhere to go, and dropping the EVM entries would leave every EVM pair unquotable against the API as it stands. Needs a Swapter-side change.

src/swap/central/swapter.ts
j0ntz: Swapter's create accepts a deposit amount only, so an exact-receive quote has nothing to send. The `quoteFor !== 'from'` guard carries that reason inline and the PR description now lists reverse quotes under known limitations; a `to` path needs the adapter to accept a withdraw amount.

src/swap/central/swapter.ts
j0ntz: Fixed. Added `addressTypeMap = { zcash: 'transparentAddress' }` and passed it into `getAddress` for both wallets, matching `changenow`/`letsexchange`. A new unit case builds a Zcash wallet that lists its unified address first and asserts the create body's `withdraw.address` is the transparent one.

src/swap/central/swapter.ts
j0ntz: The single `{ error: { code, message, min?, max? } }` object is what the adapter returns, and the plugin cannot synthesize an `errors` array it never receives. Needs a Swapter-side change.

src/mappings/swapter.ts
j0ntz: Fixed, each confirmed against live `/data/coins` and `deposit-range` today. `EOS`/`EOS` is rejected with code 1 while `A`/`EOS` quotes (min 501.39), so `eos` gets a `SPECIAL_MAINNET_CASES` entry for Vaulta's `A`. `ETH`/`ZKS20` is rejected while `ZK`/`ZKS20` quotes, so the network stays mapped and only the native is blocked, via `INVALID_TOKEN_IDS: { zksync: [null] }`. `RVN` is absent from the live list, so `ravencoin` is now `null`. `swapterMap.json` was regenerated from today's snapshot: 44 networks / 829 tickers, up from 38 / 729.

src/swap/central/swapter.ts
j0ntz: Fixed three ways. `lastUpdated` is stamped only on a non-empty result, so an empty success is retried on the next quote instead of being cached for the hour. The quote now fails with a provider error when the map is empty rather than falling through to `getChainAndTokenCodes`, so an outage no longer reads as an unsupported pair. And the cache moved inside `makeSwapterPlugin`, making it per-instance so one test can no longer stamp the next; the fake `/data/coins` now returns real rows, with a new case covering the empty-map path.

src/swap/central/swapter.ts
j0ntz: Fixed. Both codes map regardless of whether Swapter echoed the bound, since core's limit errors accept `undefined`, so a bound-less rejection stays pair-level and the float fallback does not retry it. A new unit case asserts exactly one create call for that shape.

src/mappings/swapter.ts
j0ntz: Fixed. The header example drops the `X-API-KEY` and states the endpoint must be called unauthenticated, with the 401-on-any-key behavior spelled out so the curl matches what `publicHeaders` does.

src/mappings/swapter.ts
j0ntz: Not verifiable from the client, and there is nothing the plugin can send to bundle an activation, so this is a question for Swapter rather than a code change. Left mapped and carried as an open item.

src/mappings/swapter.ts
j0ntz: Fixed, `axelar` now maps to `AXL`. Live `deposit-range` quotes it in both directions today: `AXL`/`AXL` to LTC min 887.80, LTC to `AXL`/`AXL` min 0.719.

src/swap/central/swapter.ts
j0ntz: Fixed, the signature is now `fetchSwapQuote(req)`.

src/swap/central/swapter.ts
j0ntz: Fixed, the ticker-map TTL is now `ASSET_CACHE_MS` and `EXPIRATION_MS` keeps the quote bound.

src/swap/central/swapter.ts
j0ntz: Fixed, the comment now refers to `deposit-range` and create rather than `min-amount`, and names native TON to `GRAM` alongside Bitcoin SV to `BCHSV`. The mapping file header carries the same list.

j0ntz:
## 📸 Test evidence (DOGE to LTC max swap)

**r5 01 swapter quote**

r5 01 swapter quote

**r5 02 swap success**

r5 02 swap success

**r5 03 swap tx recorded**

r5 03 swap tx recorded

paullinator (approved):

PR #478, opened 2026-07-23
Fix NYM max swaps from EVM wallets

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

Fixes [Swap (NYM) - Max amount reports unsupported route (for tokens)](https://app.asana.com/0/1215088146871429/1216815078906704).

Tapping MAX on the Exchange scene with an Ethereum wallet as the source (USDT on Ethereum to native NYM) failed with "No enabled exchanges support USDT to NYM", while typing the same amount manually quoted fine through NYM. The follow-up verification confirmed the same "no enabled exchanges" message reproduces from a native ETH wallet, not just tokens. Two distinct defects both surface as that misleading message because edge-core-js ranks the real error below an unrelated plugin's `SwapCurrencyError`.

**Root cause 1 (max-swap spend to self).** NYM's max-swap path builds a fee-estimation probe (`fetchProbeOrder`) so a max request does not create and abandon a live NYM order. `getMaxSwappable` hands that probe's `spendInfo` to `getMaxSpendable`, and the probe's only available from-chain address is the user's own refund address. `makeSpendCheck` in edge-currency-accountbased throws `SpendToSelfError` whenever a spend target equals `walletLocalData.publicKey`, and on every EVM chain that public key IS the address, so the probe threw before it could price anything. `SpendToSelfError` is not one of the swap error types edge-core-js ranks, so `pickBestError` preferred the `SwapCurrencyError` raised by the plugins that genuinely do not support NYM. This affects both tokens and native ETH (the public-key-is-address property is the same); native NYM as the source was unaffected because Cosmos stores a base16 pubkey while its addresses are bech32, so the self-check never matched. That matches the reporter's note that "NYM itself is able to max exchange and works".

**Fix 1.** Mark the probe spend `skipChecks`. It is never broadcast and exists only so `getMaxSpendable` can price the network fee before the real order (and its payin address) exists. The real order path keeps every check.

**Root cause 2 (out-of-limit amount).** NYM rejects an amount outside an asset's min/max with an HTTP 400 (`UnderLimitError` / `OverLimitError`) BEFORE it returns a quote, so the plugin never reaches the min/max check that a successful quote runs. `handleErrorResponse` mapped every error-array body to `SwapCurrencyError`, which again loses error ranking to other plugins. In the app this is what a native ETH wallet holding less than NYM's 0.005 ETH minimum hits on a max swap: the whole-balance probe gets `UnderLimitError`, mismapped to `SwapCurrencyError`, and the user sees "No enabled exchanges support ETH to NYM" instead of a below-limit message.

**Fix 2.** Map NYM's `UnderLimitError` / `OverLimitError` to the ranked `SwapBelowLimitError` / `SwapAboveLimitError`, carrying NYM's own native-unit boundary and following the request direction (a `to` quote is bounded by the destination amount, everything else by the source amount). These rank above `SwapCurrencyError`, so the correct "below the limit" / "above the limit" message wins.

**Same-wallet vs different-wallet note.** NYM provides its own liquidity, so its API requires native NYM on one side: an EVM-to-EVM pair (the "same EVM wallet on both sides" case) is rejected by NYM itself with "Unsupported swap pair: one side must be NYM", and the plugin gates that locally. A NYM swap is therefore always an EVM source wallet to a separate NYM wallet; the self-spend probe targets the source wallet's own refund address regardless of the destination, which is why the bug did not depend on the destination wallet.

`test/nym.test.ts` covers all of it: fake wallets mirror the engine's `SpendToSelfError` guard, and the tests assert a `quoteFor: 'max'` request quotes the full balance for both an EVM token and a native EVM asset (probe carries `skipChecks`, real order does not), that an EVM-to-EVM pair is rejected, and that `UnderLimitError` / `OverLimitError` map to `SwapBelowLimitError` / `SwapAboveLimitError` with NYM's limit. The self-spend and limit tests fail on the pre-fix code.

**Testing.** Verified in-app on the iOS simulator (`edge-funds`) with this branch's plugin bundle installed, default provider set enabled:
- USDT (Ethereum, 32.468) in My Ether 4 to native NYM, MAX: now resolves to a full NYM quote (1,901.66 NYM, "Slide to Confirm") instead of the exchange error.
- Native ETH (My Ether 4, 0.0035 ETH, below NYM's 0.005 min) to native NYM, MAX: now shows "Amount is below the min limit of 0.005 ETH" instead of "No enabled exchanges support ETH to NYM".

`npm run verify` (prepare + tsc + eslint + mocha) passes. Screenshots in the comment below.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - fixed: (NYM) Max swaps from EVM wallets no longer fail with "No enabled exchanges support". The fee-estimation probe targets the user's own address, which the engine rejected as a spend to self.
+- fixed: (NYM) Swaps whose amount is outside NYM's per-asset limits now report "below the limit" / "above the limit" instead of "No enabled exchanges support". NYM rejects an out-of-range amount before quoting, and the error was mapped to an unsupported-pair error that lost error ranking to other plugins.
 
 ## 2.52.0 (2026-07-17)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- fixed: (NYM) Max swaps from EVM wallets no longer fail with "No enabled exchanges support". The fee-estimation probe targets the user's own address, which the engine rejected as a spend to self.
+
 ## 2.52.0 (2026-07-17)
 
 - added: (Bridgeless) Solana (native SOL) and TON (native) swaps to and from Zano, including referral id support.
--- a/src/swap/central/nym.ts
+++ b/src/swap/central/nym.ts
@@ -303,6 +303,14 @@ export function makeNymPlugin(opts: EdgeCorePluginOptions): EdgeSwapPlugin {
         }
       ],
       networkFeeOption: 'high',
+      // This spend is never broadcast: it exists only so `getMaxSpendable` can
+      // price the network fee before the real order (and its payin address)
+      // exists. Its target is the user's own from-chain address, which engines
+      // that compare the spend target against their own public key reject with
+      // `SpendToSelfError` (every EVM chain, where the public key *is* the
+      // address). That threw out of `getMaxSwappable` and failed every max
+      // swap from an EVM wallet. The real order below keeps all checks.
+      skipChecks: true,
       assetAction: {
         assetActionType: 'swap'
       }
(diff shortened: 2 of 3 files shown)


j0ntz:
## 📸 Test evidence

**before fix max exchange error**

before fix max exchange error

**max pair usdt eth to nym**

max pair usdt eth to nym

**max quote succeeds**

max quote succeeds

peachbits (approved):

j0ntz:
## 📸 Test evidence (follow-up: native ETH + limit-error fix)

**usdt token max succeeds**

usdt token max succeeds

**eth native below limit**

eth native below limit

**eth native before fix no exchanges**

eth native before fix no exchanges

PR #480, opened 2026-07-31
Add MoonPay Trade swap provider

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

Adds MoonPay Trade, MoonPay's swap API, as a swap provider, routing EVM, Solana and deposit-address chains through one shared factory registered twice: `mptrade`, displayed as "MoonPay Trade (Centralized)", and `mptradedefi`, displayed as "MoonPay Trade (DeFi)" (Solana-to-Solana only).

Asana Task: https://app.asana.com/0/1215088146871429/1217036054017879

**Route models.** `getAction`'s payload shape depends on the SOURCE chain's `vmId`, so each gets its own cleaner and spend builder:

- `evm`: router calldata. `data` is `'0x'` on a plain value send to a bridge contract, which must not become an empty hex memo.
- `solana`: an unsigned v0 `VersionedTransaction` handed to the engine through `otherParams.unsignedTx`.
- `alt-vm`: a deposit address on the source chain, with `toExtra` carrying the memo or tag when the chain needs one.

`hypercore` and any non-`DEFAULT` execution type are rejected. 26 non-EVM chains are mapped from a live `getChainList`, keyed on the reported `vmId` rather than on whether the chain id looks EVM: 314, 10000, 2000 are all real EVM chain ids AND all tagged `alt-vm`, confirmed by the deposit-address format each returns.

**Order registration.** Routes flagged `requiresRegisterTransaction` POST the broadcast hash back to MoonPay Trade, wrapping the quote's `approve` since `makeSwapPluginQuote` has no post-broadcast hook. A registration failure is logged and swallowed: the swap is already on chain by then, so throwing would report a successful swap as failed.

**Fixed-rate quotes.** The quote publishes `amountOutMin`, the route's on-chain floor, and is marked `isEstimate: false`. The user may receive more and can never receive less, which is what a fixed quote asserts. This is MoonPay Trade's own recommendation over quoting the expected `amountOut` and calling the route an estimate. `minReceiveAmount` is left unset because the quoted amount IS the minimum.

**Tiered slippage.** 10 bps when both legs are stablecoins, 50 bps when both are stable or major, 100 bps otherwise; a route takes its looser leg. Because the quote publishes the floor, a tighter band directly raises the amount the user is guaranteed. The tiers are overridable at runtime from the info server through `opts.infoPayload.slippageBps`, so they can be retuned per asset class with no app release.

**Venue split (two registrations).** `isDex` lives on `EdgeSwapInfo`, which the core copies onto every quote, so classification is per plugin rather than per route. MoonPay Trade's route families sit on different venues, so the provider ships as two registrations that share one factory (`makeMpTradeBasedPlugin`) and partition every pair with a single predicate, so no route is quoted twice or dropped:

- `mptrade` (`central/mptrade.ts`, `isDex: false`, "MoonPay Trade (Centralized)"): every route a server can gate. EVM payloads carry a MoonPay Trade server signature the router requires, `alt-vm` sources pay an operator-issued deposit address, and every cross-chain route settles through a bridge set that is mixed on custody (mayan / across / alt-vm-3 / ghost-v2 consensus-enforced; relay / alt-vm-1/2 operator-run, per the partner).
- `mptradedefi` (`defi/mptradeDefi.ts`, `isDex: true`, "MoonPay Trade (DeFi)"): Solana-to-Solana only. The partner confirmed MoonPay Trade invokes the underlying Solana program directly with no router of its own, that no order state depends on user identity, and that `registerTransaction` is tracking only; delivery is inside the user's own atomic transaction, so nothing can gate settlement after signing.

Both entries carry the MoonPay Trade brand and their display names say which venue each covers, because Swap Settings and the preferred-provider picker show `displayName` alone. The pluginIds are `mptrade` / `mptradedefi`; the chain mapping (`src/mappings/mptrade.ts`), the info-server `slippageBps` override key and the CDN icon paths (`/exchangeIcons/mptrade/`, `/exchangeIcons/mptradedefi/`) key on them.

**Testing.** One real in-app swap per route model driven to the `Congratulations!` scene on the iOS sim: EVM source into an alt-vm destination, an alt-vm source (deposit spend plus the registration callback), a Solana source, and Base USDC to Arbitrum USDC for the fixed-quote change. That last one quoted 3.493 USDC with no variable marker, sat under the Fixed Quotes section, and delivered 3.494 USDC against the 3.493 floor. For the venue split, SOL to USDC on Solana quoted through the DeFi entry labeled DEX at a 1.493 USDC floor and delivered 1.501804 USDC in the same atomic transaction, while SOL to Base USDC still quoted through the Centralized entry. After the rename, the same Solana pair quoted as "Powered by MoonPay Trade (DeFi)", the provider sheet listed "MoonPay Trade (DeFi)" labeled DEX under both Selected Quote and Fixed Quotes, a real 0.014415 SOL to USDC (Solana) swap reached the Congratulations scene with Exchange Details naming the service "MoonPay Trade (DeFi)", and a USDC (Solana) to USDC (Base) route quoted as "Powered by MoonPay Trade (Centralized)". With the uploaded partner icon live on the CDN, a USDC to SOL (Solana) quote shows the MoonPay icon in the "Powered by MoonPay Trade (DeFi)" card (frame 42; both pluginIds serve the same icon file). Unit suite covers both quote shapes, every slippage tier, and the partition from both sides; screenshots are in the Test evidence table below.

---

> [!NOTE]
> **Medium Risk**
> New swap execution paths move user funds across EVM, Solana, and deposit-address chains with fixed guaranteed amounts; mistakes in spend construction or limit scaling would have direct financial impact, though the change is additive with broad unit coverage.
> 
> **Overview**
> Adds **MoonPay Trade** as a new cross-chain swap provider backed by the swaps.xyz API, wired into the plugin bundle as **`mptrade`** and **`mptradedefi`**.
> 
> A shared quote factory drives both registrations: it maps Edge chains via `src/mappings/mptrade.ts`, pre-checks routes and **decimal** min/max limits through `getPaths`, then builds spends from `getAction` for **EVM** (router calldata plus optional ERC-20 approvals), **Solana** (`unsignedTx`), and **alt-vm** (deposit address, including memo/tags). Quotes are **fixed** using `amountOutMin` (`isEstimate: false`), with tiered slippage (stable/major/default, overridable from the info server). Routes that require it POST the broadcast tx hash via a wrapped `approve`. **Solana-to-Solana** pairs go only through the DeFi registration; all other routes use the centralized one so each pair is quoted once.
> 
> The unreleased CHANGELOG notes the addition, and **`test/mptrade.test.ts`** covers spend shaping, limits, errors, registration, the venue split, and slippage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cc052f8629191f4e6885bfbea626300a466dde0d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### Test evidence

<table>
<tr>
<td rowspan="11" width="170" valign="top"><code>(earlier)</code><br><sub>Add swaps.xyz swap provider</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260730-185043-agent-proof-1217036054017879-01-swapsxyz-quote.png"></a><br><sub><b>1.</b> swapsxyz quote</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260730-185043-agent-proof-1217036054017879-02-swapsxyz-success.png"></a><br><sub><b>2.</b> swapsxyz success</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260803-163702-agent-proof-1217036054017879-01-swapsxyz-quote.png"></a><br><sub><b>1.</b> swapsxyz quote</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260803-163702-agent-proof-1217036054017879-02-swap-success.png"></a><br><sub><b>2.</b> swap success</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260803-163702-agent-proof-1217036054017879-03-tx-details.png"></a><br><sub><b>3.</b> tx details</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260803-163702-agent-proof-1217036054017879-04-tx-list-row.png"></a><br><sub><b>4.</b> tx list row</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260804-120242-agent-proof-1217036054017879-05-sonic-quote-decimal-limits.png"></a><br><sub><b>5.</b> sonic quote decimal limits</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260804-120242-agent-proof-1217036054017879-06-sonic-swap-success.png"></a><br><sub><b>6.</b> sonic swap success</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260804-182142-agent-proof-1217036054017879-07-evm-to-altvm-quote.png"></a><br><sub><b>7.</b> evm to altvm quote</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260804-182142-agent-proof-1217036054017879-08-evm-to-altvm-success.png"></a><br><sub><b>8.</b> evm to altvm success</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260804-182142-agent-proof-1217036054017879-09-altvm-source-quote.png"></a><br><sub><b>9.</b> altvm source quote</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260804-182142-agent-proof-1217036054017879-10-altvm-source-success.png"></a><br><sub><b>10.</b> altvm source success</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260804-182142-agent-proof-1217036054017879-11-solana-quote.png"></a><br><sub><b>11.</b> solana quote</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260804-182142-agent-proof-1217036054017879-12-solana-success.png"></a><br><sub><b>12.</b> solana success</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260804-184026-agent-proof-1217036054017879-13-retest-after-review-fixes.png"></a><br><sub><b>13.</b> retest after review fixes</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260819-105747-agent-proof-1217036054017879-14-fixed-quote-no-tilde.png"></a><br><sub><b>14.</b> fixed quote no tilde</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260819-105747-agent-proof-1217036054017879-15-fixed-quotes-section.png"></a><br><sub><b>15.</b> fixed quotes section</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260819-105747-agent-proof-1217036054017879-16-swap-success.png"></a><br><sub><b>16.</b> swap success</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260819-105747-agent-proof-1217036054017879-17-delivered-balance.png"></a><br><sub><b>17.</b> delivered balance</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260901-130058-agent-proof-1217036054017879-18-solana-dex-quote.png"></a><br><sub><b>18.</b> solana dex quote</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260901-130058-agent-proof-1217036054017879-19-provider-sheet-dex-label.png"></a><br><sub><b>19.</b> provider sheet dex label</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260901-130058-agent-proof-1217036054017879-20-solana-dex-swap-success.png"></a><br><sub><b>20.</b> solana dex swap success</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260901-130058-agent-proof-1217036054017879-21-solana-crosschain-stays-centralized.png"></a><br><sub><b>21.</b> solana crosschain stays centralized</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260908-102444-agent-proof-1217036054017879-22-defi-quote-moonpay-trade-name.png"></a><br><sub><b>22.</b> defi quote moonpay trade name</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260908-102444-agent-proof-1217036054017879-23-provider-sheet-defi-label.png"></a><br><sub><b>23.</b> provider sheet defi label</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260908-102444-agent-proof-1217036054017879-24-defi-swap-success.png"></a><br><sub><b>24.</b> defi swap success</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260908-102444-agent-proof-1217036054017879-25-exchange-details-service-name.png"></a><br><sub><b>25.</b> exchange details service name</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260908-102444-agent-proof-1217036054017879-26-centralized-quote-moonpay-trade-name.png"></a><br><sub><b>26.</b> centralized quote moonpay trade name</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260915-163852-agent-proof-1217036054017879-30-defi-quote-after-rebase.png"></a><br><sub><b>30.</b> defi quote after rebase</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260915-163852-agent-proof-1217036054017879-31-defi-swap-success-after-rebase.png"></a><br><sub><b>31.</b> defi swap success after rebase</sub></td>
</tr>
<tr>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260916-175854-agent-proof-1217036054017879-40-mptradedefi-quote-after-rename.png"></a><br><sub><b>40.</b> mptradedefi quote after rename</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260916-175854-agent-proof-1217036054017879-41-mptradedefi-swap-success-after-rename.png"></a><br><sub><b>41.</b> mptradedefi swap success after rename</sub></td>
<td width="210"></td>
</tr>
<tr>
<td rowspan="1" width="170" valign="top"><a href="https://github.com/EdgeApp/edge-exchange-plugins/pull/480/commits/cc052f8629191f4e6885bfbea626300a466dde0d"><code>cc052f8</code></a><br><sub>Split MoonPay Trade DEX routes into a second plugin</sub><br><sub>2026-09-18</sub></td>
<td width="210" valign="top" align="center"><a href="https://raw.githubusercontent.com/EdgeApp/edge-dev-agents/agent-pr-assets/assets/edge-exchange-plugins/pr-480/20260918-120634-agent-proof-1217036054017879-42-mptradedefi-powered-by-icon.png"></a><br><sub><b>42.</b> mptradedefi powered by icon</sub></td>
<td width="210"></td>
<td width="210"></td>
</tr>
</table>


j0ntz:
## 📸 Test evidence (iOS sim — real swaps.xyz swap)

**swapsxyz quote**

swapsxyz quote

**swapsxyz success**

swapsxyz success

j0ntz:
bugbot run

src/mappings/swapsxyz.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * The value is the decimal chain id as a string; it is parsed back to a number
+ * for the `srcChainId`/`dstChainId` query params.
+ *
+ * See https://docs.swaps.xyz/ for the API docs.
+ */
+
+import { EdgeCurrencyPluginId } from '../util/edgeCurrencyPluginIds'
+
+export const swapsxyz = new Map<EdgeCurrencyPluginId, string | null>()
```
peachbits, line 40: Let's add all the EVM chains from accountbased for futureproofing

src/swap/central/swapsxyz.ts
```diff
@@ -0,0 +1,390 @@
...
+      srcChainId: fromChainId,
+      srcToken: fromTokenAddress,
+      dstChainId: toChainId,
+      dstToken: toTokenAddress,
+      amount: nativeAmount,
+      swapDirection: 'exact-amount-in',
+      slippage: SLIPPAGE_BPS
+    })
+
+    const response = await fetchCors(
```
peachbits, line 560: Let's call the https://docs.swaps.xyz/swap-api-reference/get-paths endpoints before getAction. The endpoint returns usable min and max amounts to throw those errors early if we detect the swap is out of bounds.

peachbits (changes requested):

src/mappings/swapsxyz.ts
j0ntz: Mapped every EVM chain edge-currency-accountbased ships, 27 in total, each with its real numeric chain id: abstract, arbitrum, avalanche, base, binancesmartchain, bobevm, botanix, celo, ethereum, ethereumclassic, ethereumpow, fantom, filecoinfevm, hyperevm, monad, opbnb, optimism, polygon, pulsechain, rsk, sonic, zksync. The five testnet/dev chains (amoy, ethDev, filecoinfevmcalibration, holesky, sepolia) are explicitly `null`, since swaps.xyz is mainnet-only and mapping them would only buy a round trip before the same rejection.

The chain id is a property of the chain, not of swaps.xyz's support for it, so this is safe now that live support is decided per quote by the `getPaths` pre-check in the other thread: an unrouted chain returns HTTP 200 with `paths: []` and becomes a `SwapCurrencyError`. Verified against the live API: `srcChainId=250` (fantom, mapped but not routed today) returns `paths: []`.

src/swap/central/swapsxyz.ts
j0ntz: Added a `GET /getPaths` call before `getAction`. It now rejects, before any route is built:

- no route for the pair (`paths: []`, which is how swaps.xyz reports an unsupported chain or token) → `SwapCurrencyError`
- a route that cannot do exact-amount-in → `SwapCurrencyError`
- `nativeAmount` under `minAmount` → `SwapBelowLimitError` carrying the minimum
- `nativeAmount` over `maxAmount` → `SwapAboveLimitError` carrying the maximum

Limits come from `paths[].amountLimits` and fall back to `srcToken.minAmount`/`maxAmount`. Per the API reference both are base-unit strings on the source token, so they compare directly against `nativeAmount` with no decimal conversion. Six unit tests cover these paths, and the response cleaner was run against live responses for base→ethereum, ethereum→hyperevm, base USDC→arbitrum USDC, and the unrouted fantom→ethereum case.

j0ntz:
## 📸 Test evidence (after review fixes): swaps.xyz swap executed on the iOS sim

**swapsxyz quote**

swapsxyz quote

**swap success**

swap success

**tx details**

tx details

**tx list row**

tx list row

src/swap/central/swapsxyz.ts
```diff
@@ -0,0 +1,459 @@
...
+
+    // Route limits win over the source token's own limits when both exist.
+    const minAmount = path.amountLimits?.minAmount ?? srcToken.minAmount
+    const maxAmount = path.amountLimits?.maxAmount ?? srcToken.maxAmount
+    if (minAmount != null && lt(nativeAmount, minAmount)) {
+      throw new SwapBelowLimitError(swapInfo, minAmount, 'from')
+    }
+    if (maxAmount != null && gt(nativeAmount, maxAmount)) {
+      throw new SwapAboveLimitError(swapInfo, maxAmount, 'from')
+    }
```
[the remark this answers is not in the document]
j0ntz: Valid, fixed. A `max` quote reaches `fetchSwapQuoteInner` as a `from` quote for the whole balance, so an above-ceiling balance was throwing instead of quoting at the ceiling.

`fetchSwapQuoteInner` now takes `isMaxRequest`, threaded from `fetchSwapQuote` for both max paths (the token branch and the `getMaxSwappable` probe). When the amount exceeds `maxAmount`: an explicit request still throws `SwapAboveLimitError`, a max request clamps to `maxAmount` and quotes there. The clamped value is what goes to `getAction` and what the order reports as `fromNativeAmount`. Covered by a new unit test (balance 10 USDC against a 5 USDC route ceiling → quote at 5000000).

src/swap/central/swapsxyz.ts
```diff
@@ -0,0 +1,468 @@
...
+ */
+const asSwapsXyzAmountLimits = asObject({
+  minAmount: asOptional(asString),
+  maxAmount: asOptional(asString)
+})
+
+const asSwapsXyzPath = asObject({
+  chainId: asNumber,
+  supportsExactAmountIn: asOptional(asBoolean, true),
+  amountLimits: asOptional(asSwapsXyzAmountLimits)
```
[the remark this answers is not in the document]
j0ntz: False positive, rejected. `asOptional` in this `cleaners` version treats JSON `null` as absent, not as a value to pass through to `asString`.

Checked directly against the installed package:

```
asObject({ minAmount: asOptional(asString), maxAmount: asOptional(asString) })
  {"minAmount":null,"maxAmount":null}  ->  {}
  {}                                   ->  {}
  {"minAmount":"5","maxAmount":null}   ->  {"minAmount":"5"}
```

The same cleaner also parsed four live `getPaths` responses whose limits are all `null` (base→ethereum, ethereum→hyperevm, base USDC→arbitrum USDC, fantom→ethereum), and a real Base USDC→ETH swap executed through this exact path in-app. A null `amountLimits` object behaves the same way and is read with `path.amountLimits?.minAmount`.

src/swap/central/swapsxyz.ts
```diff
@@ -0,0 +1,468 @@
...
+    throw new SwapBelowLimitError(swapInfo, undefined, 'from')
+  }
+  if (
+    upper.includes('MAX') ||
+    upper.includes('HIGH') ||
+    upper.includes('LARGE') ||
+    upper.includes('EXCEED')
+  ) {
+    throw new SwapAboveLimitError(swapInfo, undefined, 'from')
+  }
```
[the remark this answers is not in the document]
j0ntz: Valid, fixed. `throwSwapsXyzError` now checks the limit keywords BEFORE the currency keywords, so a limit failure that also names the route or token ranks as the limit error rather than `SwapCurrencyError`.

The keywords are whole phrases now instead of bare substrings, which removes the `ALLOWANCE` contains `LOW` class of misclassification: below = TOO_LOW / TOO LOW / TOO_SMALL / TOO SMALL / MINIMUM / BELOW, above = TOO_HIGH / TOO HIGH / TOO_LARGE / TOO LARGE / MAXIMUM / EXCEED / ABOVE. Two new tests cover it: `NO_QUOTE` + "Amount too low for this route" now ranks as `SwapBelowLimitError`, and `ALLOWANCE_REQUIRED` stays a plain `Error`. The live dust case still classifies correctly (`AMOUNT_TOO_LOW` → `SwapBelowLimitError`).

src/swap/defi/swapsxyz.ts
```diff
@@ -0,0 +1,468 @@
...
+    if (action.amountOut.amount === '0' || action.amountOutMin.amount === '0') {
+      throw new SwapBelowLimitError(swapInfo, undefined, 'from')
+    }
+
+    // ERC20 sources must approve the router (`tx.to`) before the swap.
+    const preTxs: EdgeTransaction[] = []
+    if (fromTokenId != null && action.requiresTokenApproval) {
+      const approvalTxs = await createEvmApprovalEdgeTransactions({
+        request,
+        approvalAmount: action.amountIn.amount,
```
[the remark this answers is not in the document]
j0ntz: Valid, fixed. The quote path now refuses any response that would authorize more of the source asset than was requested:

```ts
if (gt(action.amountIn.amount, swapAmount) || gt(action.tx.value, swapAmount)) {
  throw new Error('swaps.xyz getAction returned a source amount above the requested amount')
}
```

`amountIn.amount` is what drives the ERC20 approval and `tx.value` the native spend, so both are bounded by the locally computed `swapAmount` before any spendInfo is built. A new unit test asserts the rejection. Verified the guard does not reject legitimate quotes: against the live API, base ETH→ethereum ETH, base USDC→ethereum ETH (token + approval path) and a `quoteFor: 'max'` request all still return quotes, with `amountIn` and `tx.value` equal to the requested amount.

src/swap/central/swapsxyz.ts
```diff
@@ -0,0 +1,492 @@
...
+    // The spend and the ERC20 approval below are built from response fields, so
+    // never let the response authorize MORE of the source asset than was asked
+    // for: `amountIn` drives the approval, `tx.value` the native spend.
+    if (
+      gt(action.amountIn.amount, swapAmount) ||
+      gt(action.tx.value, swapAmount)
+    ) {
+      throw new Error(
+        'swaps.xyz getAction returned a source amount above the requested amount'
+      )
```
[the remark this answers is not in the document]
j0ntz: Valid on the unit mismatch, fixed. The guard now compares `tx.value` against `swapAmount` ONLY when the source is the native asset; on a token route `amountIn.amount` (source-token units) is the sole comparison, so a native protocol fee in `tx.value` can no longer falsely reject a quote. New unit test: a token quote with `tx.value` of 30000000000000 wei and `amountIn` of 3000000 USDC base units returns a quote.

On the second half, `makeSwapsXyzSpendInfo` dropping `tx.value` for token routes: that predates this PR and is not something I want to change blind. A single `EdgeSpendInfo` spendTarget carries one asset, so attaching a native fee alongside an ERC20 spend needs a decision about how it is expressed, and every swaps.xyz token route observed so far returns `tx.value` of 0 (verified against the live API for base USDC→ethereum ETH). Recorded as a follow-up rather than guessed at here.

src/swap/central/swapsxyz.ts
j0ntz: Correction to the reply above: the limits are NOT base units. `getPaths` returns `minAmount`/`maxAmount` as DECIMAL strings on the source token. The swaps.xyz API reference says base units and is wrong, and every pair used to build and test the original check happened to return null limits, so the mistake shipped unseen.

Proven against the live API: `getPaths` for ethereum -> sonic returns `decimals: 18`, `minAmount: '0.0001325'`, `maxAmount: '6.404342'`. Read as base units, a 0.25 ETH request compared `250000000000000000` against `6.404342` and threw `SwapAboveLimitError`. Every route that publishes limits was affected; today that is a minority of EVM pairs (ethereum -> base, arbitrum -> ethereum and polygon -> ethereum all return null), which is why it went unnoticed.

Fixed: `asSwapsXyzPaths` now cleans `srcToken.decimals`, and a `limitToNative` helper scales each limit into base units before the compare (rounding a floor limit up and a ceiling limit down, so a sub-unit residue on a low-decimal token cannot widen the allowed range). The scaled value is what the error carries, so `nativeMin`/`nativeMax` reach the GUI in the units it formats.

Re-verified live after the fix: ethereum -> sonic 0.25 ETH now quotes; 10 ETH throws `SwapAboveLimitError` with `nativeMax` 6404342000000000000; 0.00001 ETH throws `SwapBelowLimitError` with `nativeMin` 132500000000000; ethereum -> base 0.25 ETH still quotes. Unit tests now use decimal limits throughout, plus three new cases: an in-range quote under a published ceiling, a 6-decimal source token, and the base-unit value on the error.

j0ntz:
## 📸 Test evidence (decimal-limit fix, limit-publishing route)

**sonic quote decimal limits**

sonic quote decimal limits

**sonic swap success**

sonic swap success

j0ntz:
## 📸 Test evidence: one in-app swap per route model (CEX reclass + solana/alt-vm dispatch)

**evm to altvm quote**

evm to altvm quote

**evm to altvm success**

evm to altvm success

**altvm source quote**

altvm source quote

**altvm source success**

altvm source success

**solana quote**

solana quote

**solana success**

solana success

j0ntz:
## 📸 Test evidence (after review fixes)

**retest after review fixes**

retest after review fixes

src/swap/central/swapsxyz.ts
```diff
@@ -0,0 +1,691 @@
...
+      // The engine executes `unsignedTx`; this address only labels the spend.
+      publicAddress = amountIn.isNative
+        ? SOLANA_SYSTEM_PROGRAM_ID
+        : amountIn.address
+      otherParams = { unsignedTx: tx.base64Tx }
+      break
+    }
+    case 'alt-vm': {
+      const tx = asSwapsXyzAltVmTx(action.tx)
+      fromNativeAmount = tx.value
```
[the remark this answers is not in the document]
j0ntz: Valid, fixed. The over-request guard is now a `rejectOverRequest` helper applied to the field each route model actually SPENDS, not just to `amountIn`: every model checks `amountIn.amount`, an `alt-vm` route additionally checks `tx.value` (the deposit amount), and a native EVM route additionally checks `tx.value` as before. A response that inflates `tx.value` past the request on an alt-vm route now throws before any spend is built.

New unit test: an alt-vm response with `amountIn` 100000000 and `tx.value` 900000000 against a 100000000 request fails with 'above the requested amount'.

src/swap/central/swapsxyz.ts
```diff
@@ -0,0 +1,691 @@
...
+  tx: asUnknown,
+  txId: asString,
+  amountIn: asSwapsXyzAmount,
+  amountOut: asSwapsXyzAmount,
+  amountOutMin: asSwapsXyzAmount,
+  vmId: asString,
+  requiresTokenApproval: asBoolean,
+  // Registration is how swaps.xyz starts tracking an order it did not itself
+  // broadcast. Their docs call it mandatory for non-EVM transactions.
+  requiresRegisterTransaction: asOptional(asBoolean, false),
```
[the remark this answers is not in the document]
j0ntz: False positive, rejected. The live `getAction` response field is `requiresRegisterTransaction`, which is what the cleaner reads.

Verified against the live API (2026-08-05, real key). The top-level keys of a `getAction` response are:

```
['tx', 'txId', 'amountIn', 'amountInMax', 'amountOutMin', 'amountOut', 'protocolFee',
 'applicationFee', 'bridgeFee', 'bridgeIds', 'bridgeRoute', 'exchangeRate',
 'estimatedTxTime', 'estimatedPriceImpact', 'vmId', 'requiresTokenApproval',
 'requiresRegisterTransaction', 'executionsType', 'allRoutes']
```

There is no `requiresRegisterTx` in the payload. The flag reads `true` on solana and alt-vm routes, and the wrap was exercised end to end: an in-app LTC (alt-vm) to Base USDC swap executed to the success scene, and a direct POST of the exact body the plugin sends returns `[{"txId":"0x9c81a482…","success":true}]` with HTTP 200.

src/mappings/swapsxyz.ts
```diff
@@ -0,0 +1,93 @@
...
+swapsxyz.set('bitcoincash', '10000')
+swapsxyz.set('bitcoinsv', '999000331')
+swapsxyz.set('bobevm', '60808')
+swapsxyz.set('botanix', '3637')
+swapsxyz.set('cardano', '1816')
+swapsxyz.set('celo', '42220')
+swapsxyz.set('cosmoshub', '999000433')
+swapsxyz.set('dash', '999000416')
+swapsxyz.set('digibyte', '999000301')
+swapsxyz.set('dogecoin', '2000')
```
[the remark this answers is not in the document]
j0ntz: False positive, rejected. At swaps.xyz, 10000 and 2000 are native Bitcoin Cash and native Dogecoin, not SmartBCH and Dogechain. Their id space reuses an EVM number wherever one exists; the `vmId` from `GET /getChainList` is what says which chain it is, and both are `alt-vm`, which is the deposit-address model no EVM sibling would use.

Verified live the same way the filecoin entry was (2026-08-05, real key). `getAction` from chainId 10000 returns a CashAddr deposit address:

```
sender qzm47qz5ue99…  -> vmId alt-vm, tx.to bitcoincash:qrs4xdffa2ta2ch9k76m04xc2dmhwdam7ygvlhnn52
sender 1BvBMSEYstWe…  -> vmId alt-vm, tx.to bitcoincash:qz2pq3dfn3hdjr5ak9g49rcxphsa2twf9yscce2jmx
```

A SmartBCH route would be `vmId: evm` and would return an 0x router address. For chainId 2000, a native DOGE sender (`DH5yaieqoZN36fDVciNyRueRGvGLR3mr7L`) is accepted and only fails on `AMOUNT_TOO_LOW`, while an 0x sender is rejected with `INVALID_ADDRESS_FORMAT` — the opposite of what a Dogechain mapping would do.

The synthetic `999000xxx` ids appear only for chains that have NO EVM number at all, which is why `bitcoin` and `litecoin` carry them and these two do not.

j0ntz:
## 📸 Test evidence (fixed-rate quotes + tiered slippage)

**fixed quote no tilde**

fixed quote no tilde

**fixed quotes section**

fixed quotes section

**swap success**

swap success

**delivered balance**

delivered balance

j0ntz:
## 📸 Test evidence (Solana same-chain DEX registration)

**solana dex quote**

solana dex quote

**provider sheet dex label**

provider sheet dex label

**solana dex swap success**

solana dex swap success

**solana crosschain stays centralized**

solana crosschain stays centralized

j0ntz:
## 📸 Test evidence (MoonPay Trade rename)

**defi quote moonpay trade name**

defi quote moonpay trade name

**provider sheet defi label**

provider sheet defi label

**defi swap success**

defi swap success

**exchange details service name**

exchange details service name

**centralized quote moonpay trade name**

centralized quote moonpay trade name

src/swap/central/swapsxyz.ts
```diff
@@ -0,0 +1,862 @@
...
+
+  switch (vmId) {
+    case 'solana': {
+      const tx = asSwapsXyzSolanaTx(action.tx)
+      fromNativeAmount = amountIn.amount
+      // The engine executes `unsignedTx`; this address only labels the spend.
+      publicAddress = amountIn.isNative
+        ? SOLANA_SYSTEM_PROGRAM_ID
+        : amountIn.address
+      otherParams = { unsignedTx: tx.base64Tx }
```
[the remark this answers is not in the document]
j0ntz: This is the Solana pattern this repo already ships: rango.ts (SOLANA case) hands the aggregator's serializedMessage to the engine as otherParams.unsignedTx with nativeAmount as the label, and lifi.ts does the same with its base64 data; no plugin here decodes v0 instructions to reconcile the transfer amount, and no Solana decoder is a dependency of this package. The signed bytes are the aggregator's for every Solana DEX integration, so instruction-level verification belongs in the engine that signs, not per plugin. Accepted risk, recorded in the run report with a reopen trigger (an engine-side transfer check, or a partner simulated-balance-change field).

src/swap/central/mptrade.ts
```diff
@@ -0,0 +1,862 @@
...
+  recentBlockhash: asString,
+  payer: asString
+})
+
+const asSwapsXyzAltVmTx = asObject({
+  to: asString,
+  toExtra: asOptional(asString),
+  value: asString,
+  chainId: asNumber
+})
```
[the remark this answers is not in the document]
j0ntz: Fixed in 16dc303: toExtra now cleans with asOptionalBlank(asNumberString), the pattern swapter and letsexchange already use, so a numeric tag and the tag 0 both survive. Added a unit test asserting a numeric 0 becomes the memo '0'.

src/mappings/swapsxyz.ts
```diff
@@ -0,0 +1,93 @@
...
+swapsxyz.set('sepolia', null) // Ethereum testnet
+swapsxyz.set('solana', '1399811149')
+swapsxyz.set('sonic', '146')
+swapsxyz.set('stellar', '999000338')
+swapsxyz.set('sui', '999000938')
+swapsxyz.set('tezos', '999000358')
+swapsxyz.set('ton', '999000337')
+swapsxyz.set('tron', '728126428')
+swapsxyz.set('zcash', '999000322')
+swapsxyz.set('zksync', '324')
```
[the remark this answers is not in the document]
j0ntz: Fixed in 16dc303: the map now carries all 86 EdgeCurrencyPluginId keys, matching rango, lifi and changenow. robinhood maps to 4663 (Robinhood Chain is on the live getChainList, and master added the currency plugin after this branch opened); the other 32 are null, testnets because swaps.xyz is mainnet-only and the rest because getChainList does not list them.

src/swap/central/mptrade.ts
```diff
@@ -0,0 +1,867 @@
...
+            ...request,
+            nativeAmount: maxAmount,
+            quoteFor: 'from'
+          }
+        } else {
+          newRequest = await getMaxSwappable(
+            async r => await fetchSwapQuoteInner(r, slippageBps, true),
+            request
+          )
+        }
```
peachbits, line 1012: **Warning:** the max probe creates and abandons a provider order on deposit-address sources.

For a native `alt-vm` source (LTC, BTC, XMR, XRP…) a max swap runs `fetchSwapQuoteInner` twice: once here with the full balance, then again with the trimmed amount. Both calls hit `getAction`, which per the provider's own docs *"allocates a deposit address, builds transaction calldata, or triggers wallet-screening checks"*, so every max swap from a UTXO/alt-vm wallet leaves an unfunded order behind. `.cursor/BUGBOT.md` (`max-probe-no-order`) accepts that only when the provider has no quote-only path, and this one does: `GET /getQuote` is priced-only, takes no `sender`, and returns `vmId`.

The EVM and Solana probes legitimately need `getAction` for the calldata / unsigned tx (same as lifi and rango). For an alt-vm source the fee estimate does not depend on the destination, so the probe can skip `getAction` entirely: limits already come from `getPaths`, and a probe spend to the user's own address with `skipChecks: true` (nym's `fetchProbeOrder`) is enough for `getMaxSpendable`. Deciding "alt-vm" locally as "not EVM (`currencyInfo.evmChainId == null`) and not solana" is safe: the EVM chains the provider sources through alt-vm (sonic, ethereumpow, pulsechain) just keep today's behavior. `getQuote`'s `vmId` is the authoritative alternative if you would rather not infer.

Please also add a test through `getMaxSwappable`: `test/mptrade.test.ts` only covers the token max path (`clamps a max request…`), so the native probe → `getMaxSpendable` → second quote path has no coverage.

src/swap/central/mptrade.ts
```diff
@@ -0,0 +1,867 @@
...
+        log.warn(
+          `MoonPay Trade registerTxs failed for ${txId}: ${
+            failure?.error ?? `status ${response.status}`
+          }`
+        )
+      }
+    } catch (error: unknown) {
+      log.warn(`MoonPay Trade registerTxs threw for ${txId}: ${String(error)}`)
+    }
+  }
```
peachbits, line 672: **Warning:** registration gets one attempt, and the failure log drops the one value support would need.

The provider docs call `registerTxs` *"Mandatory for non EVM transactions"* and *"required for any transactions that return an `alt-vm-*` bridge ID"*. On a deposit-address route the user's coins are already at the provider's address when this runs, so a single transient failure (5xx, a dropped connection right after broadcast) leaves that deposit unindexed until someone registers it by hand. Swallowing is right, since throwing would report a settled swap as failed, but a simple retry loop here (about 5 attempts with a short backoff, still bounded so `approve` returns promptly) covers the transient case without any outside help. Then:

- log the final failure with **both** `txId` and `txHash`; today only `txId` is logged, and the hash is what a manual register call needs
- `response.json()` runs before `response.ok` is checked, so a non-JSON 5xx body lands in the `catch` as a parse error and the log loses the status
- add a unit test that a failing `/registerTxs` still resolves `approve()`; only the success and not-required cases are covered

src/swap/central/mptrade.ts
```diff
@@ -0,0 +1,867 @@
...
+
+  if (BELOW_LIMIT_KEYWORDS.some(keyword => upper.includes(keyword))) {
+    throw new SwapBelowLimitError(swapInfo, undefined, 'from')
+  }
+  if (ABOVE_LIMIT_KEYWORDS.some(keyword => upper.includes(keyword))) {
+    throw new SwapAboveLimitError(swapInfo, undefined, 'from')
+  }
+  if (CURRENCY_ERROR_KEYWORDS.some(keyword => upper.includes(keyword))) {
+    throw new SwapCurrencyError(swapInfo, request)
+  }
```
peachbits, line 609: **Warning:** classify on the provider's error codes, not on keywords in the message.

The spec publishes a closed `code` enum, so the mapping can be exact: `UNSUPPORTED_NETWORK`, `UNSUPPORTED_NETWORK_PAIR`, `UNSUPPORTED_NETWORK_TOKEN_PAIR`, `NO_AVAILABLE_ROUTE`, `INVALID_SOURCE_TOKEN`, `INVALID_DESTINATION_TOKEN`, `INVALID_ADDRESS_FORMAT`, `UNSUPPORTED_SWAP_DIRECTION` → `SwapCurrencyError`; `AMOUNT_TOO_LOW` / `AMOUNT_TOO_HIGH` (and `INVALID_AMOUNT_ZERO`) → the limit errors; `INSUFFICIENT_LIQUIDITY`, `INTERNAL_SERVER_ERROR` and anything unrecognised → plain `Error`.

Today the keyword scan ignores the cleaned `statusCode`, so a `getPaths` 500 with `{ code: 'INTERNAL_SERVER_ERROR', message: 'Failed to load paths for token' }` matches `PATH`/`TOKEN` and reports "unsupported pair" instead of a failed provider (`.cursor/BUGBOT.md` `outage-is-not-unsupported`). If you keep a keyword fallback for codes outside the enum, gate it on a 4xx (`swapError.error.statusCode ?? response.status`) so a 5xx always falls through to the plain `Error`.

src/swap/central/mptrade.ts
```diff
@@ -0,0 +1,867 @@
...
+  throw new Error(
+    `MoonPay Trade ${endpoint} failed: ${code}${
+      message !== '' ? ` (${message})` : ''
+    }`
+  )
```
peachbits: **Warning:** `GEO_BLOCKED` surfaces as a generic provider failure.

The published code enum includes `GEO_BLOCKED` and `WALLET_SCREENED`. Neither matches a keyword, so a geo-blocked user gets `Error('MoonPay Trade getAction failed: GEO_BLOCKED …')`. nym and `template.ts` map region failures to `SwapPermissionError(swapInfo, 'geoRestriction')`, which the GUI knows how to present; `GEO_BLOCKED` should do the same here (and `WALLET_SCREENED` is worth a deliberate mapping too, even if it stays a plain error).

src/swap/central/mptrade.ts
```diff
@@ -0,0 +1,867 @@
...
+    if (path == null || !path.supportsExactAmountIn) {
+      throw new SwapCurrencyError(swapInfo, request)
+    }
+
+    // Route limits win over the source token's own limits when both exist.
+    // Both arrive as decimal strings, so they have to be scaled into the base
+    // units `nativeAmount` uses before any compare.
+    const { decimals } = srcToken
+    const minLimit = path.amountLimits?.minAmount ?? srcToken.minAmount
+    const maxLimit = path.amountLimits?.maxAmount ?? srcToken.maxAmount
```
peachbits: **Suggestion:** the precedence here is the reverse of the spec. The OpenAPI marks `paths[].amountLimits` deprecated: *"maintained for backwards compatibility. Use `srcToken.minAmount` and `srcToken.maxAmount` at the top level of the response instead."* Live today both carry the same values (checked eth→sonic, ltc→base, base→xmr, sol→sol), so no behavior change now, but reading `srcToken` first with `amountLimits` as the fallback matches the spec, and the comment ("Route limits win…") currently states the opposite.

src/swap/central/mptrade.ts
```diff
@@ -0,0 +1,867 @@
...
+  data: asString,
+  value: asString,
+  chainId: asNumber
+})
+
+const asMpTradeSolanaTx = asObject({
+  base64Tx: asString,
+  recentBlockhash: asString,
+  payer: asString
+})
```
peachbits, line 180: **Suggestion:** `recentBlockhash` and `payer` are cleaned and never read (the engine re-fetches the blockhash at sign time). Same for `chainId` on the three `tx` cleaners, `decimals`/`symbol`/`chainId` on `asMpTradeAmount`, and `name` on `asMpTradeError`. Please either comment them out or remove them (`remove-unused-cleaner-fields`); `statusCode` becomes used if you take the classification change above.

peachbits (changes requested):
Re-review of the rebased head (cc052f8). Both earlier asks are addressed (full EVM chain map, getPaths pre-check). Local `tsc`, `eslint` and the full mocha suite (142) pass on this head.

Two git-hygiene notes that have no line to attach to:
- `Split MoonPay Trade DEX routes into a second plugin` is 51 chars; the limit is 50. Suggest `Split MoonPay Trade DEX routes into own plugin` (46).
- The body of `Add MoonPay Trade swap provider` still describes a DEX plugin that rejects non-EVM VMs. When the fixups squash into it, rewrite it for what ships (CEX-classified, evm/solana/alt-vm route models).

src/swap/central/mptrade.ts
j0ntz: Done. `fetchProbeOrder` now tests the source locally (`currencyInfo.evmChainId == null && pluginId !== 'solana'`) and, for a deposit-address source, builds a self-send spend with `skipChecks: true` and prices it through `getMaxSpendable`, so no `getAction` call happens on the probe and only the trimmed quote allocates an order. EVM and Solana sources keep the `getAction` probe, since the calldata and the unsigned transaction come from there.

The pair gating, `getPaths` call and ceiling clamp moved into a shared `fetchRoute`, so the probe still gets its limits from `getPaths`. Two tests cover it through `getMaxSwappable`: a native LTC max asserts exactly one `/getAction` plus the probe spend target, and an EVM max asserts two.

src/swap/central/mptrade.ts
j0ntz: Done. `registerTx` now retries on a bounded backoff (250/500/1000/2000 ms, five attempts, under four seconds so `approve` stays prompt), checks `response.ok` before parsing so a non-JSON 5xx keeps its status in the log, and logs the final failure with both `txId` and `txHash`.

One deliberate narrowing: a parsed `success: false` is the provider's answer about that hash rather than a transient fault, so it breaks out instead of re-POSTing, which would risk a double registration. Three tests cover it: `approve()` resolves after every attempt fails (five calls, one warning carrying the hash and `status 503`), a transient failure that lands on the retry, and a rejection that is not re-POSTed.

src/swap/central/mptrade.ts
j0ntz: Done. Classification now keys on the published enum: four code sets map to `SwapCurrencyError`, the limit errors, and plain `Error` for `INSUFFICIENT_LIQUIDITY`, `INTERNAL_SERVER_ERROR` and `WALLET_SCREENED`. The keyword scan survives only as a fallback for codes outside the enum and only on a 4xx: `throwMpTradeError` takes the HTTP status from both call sites and reads `statusCode ?? status`, so anything outside 400-499 goes straight to the plain `Error`.

A test covers the case you named: `PATHS_UNAVAILABLE` with a path/token message reads as `SwapCurrencyError` on a 400 and as a provider failure naming `getPaths` on a 500.

src/swap/central/mptrade.ts
j0ntz: Done. `GEO_BLOCKED` now throws `SwapPermissionError(swapInfo, 'geoRestriction')`, and `WALLET_SCREENED` is mapped deliberately to a plain `Error` alongside the other provider-side codes, since a screened wallet is not a region the user can leave. Both have tests.

src/swap/central/mptrade.ts
j0ntz: Fixed. `srcToken.minAmount`/`maxAmount` are read first with `path.amountLimits` as the fallback, and the comment now says which one the spec prefers and why the other is kept.

src/swap/central/mptrade.ts
j0ntz: Removed. `asMpTradeSolanaTx` is down to `base64Tx`, `chainId` is gone from the EVM and alt-vm tx cleaners, `asMpTradeAmount` keeps `amount`, `address` and `isNative`, and `asMpTradeError` keeps `code`, `message` and `statusCode`, which the new classification reads.

j0ntz:
All six inline threads are answered and resolved, with one fixup carrying the code. The suite is 151 on this head (`tsc` and `eslint` clean).

Both git-hygiene notes are owed at the squash and recorded for it, since rewording a published commit now would rewrite the history you are mid-review on: `Split MoonPay Trade DEX routes into own plugin` (46) and a body for `Add MoonPay Trade swap provider` that describes what ships (CEX-classified, evm/solana/alt-vm route models) land with the fold.

One deviation worth flagging: this round's fixup targets the split commit, not the provider commit. It rewrites the `handlesRoute` gate and the `throwMpTradeError` call sites that the split introduced, so a fold into the earlier commit cannot replay the split (`slot-fixup.sh` proved it by conflicting). The provider commit's body still gets the rewrite above.

PR #483, opened 2026-08-13
Add Robinhood Chain swap support

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

Companion PRs for the same chain (this PR is independent of both and passes CI standalone):

- [EdgeApp/edge-currency-accountbased#1087](https://github.com/EdgeApp/edge-currency-accountbased/pull/1087) adds the `robinhood` currency plugin
- [EdgeApp/edge-react-gui#6150](https://github.com/EdgeApp/edge-react-gui/pull/6150) registers it in the app

### Description

Maps the `robinhood` plugin id (Robinhood Chain, EVM chain 4663) onto each swap provider's own chain code, so quotes can route to and from the chain. No engine or plugin logic changes: `lifi.ts`, `rango.ts` and the central plugins all gate on a non-null `MAINNET_CODE_TRANSCRIPTION` lookup, so a chain the map does not know is invisible to the provider no matter what its API supports.

`robinhood` is added to `src/util/edgeCurrencyPluginIds.ts` via `mapctl add-plugin`; the six provider entries are authored in `scripts/mappings/*Mappings.ts` and the `src/mappings/*.ts` files are regenerated with `npm run mapctl update-mappings`.

Two of those generated files are hand-added here rather than regenerated. `updateMappings` in `scripts/mapctl.ts` iterates the synchronizer list, and `changelly` has a mapping file but no synchronizer while `nexchange` is hand-maintained with no source file at all, so the generator can never write either one. Both were missing exactly the `robinhood` key against every sibling, which `docs/CHAIN_MAPPING_SYNCHRONIZERS.md` says must be an explicit `null` so a later maintainer can tell a deliberate exclusion from a missed one. Making the sweep iterate the authored mapping files instead is left as follow-up work.

| Provider | Chain code | Why |
|---|---|---|
| SideShift | `robinhood` | `/v2/pair/btc-bitcoin/eth-robinhood` and `/v2/pair/eth-arbitrum/eth-robinhood` both quote, in both directions |
| LI.FI | `out` | `/v1/quote` routes cross-chain into 4663. The key really is the string `out`, which is what `/v1/chains` returns for this chain |
| Rango | `ROBINHOOD` | Chain enabled with 495 tokens; quotes via Relay and GasZip |
| ChangeNow | `hood` | Native ETH, USDG, CASHCAT and tokenized equities. The native-ETH pair was inactive when first mapped and fills now |
| Swapuz | `ROBINHOOD` | Native ETH, USDG, CASHCAT, PONS, PIPEDOG. `rate/` quotes 0.05 ETH into 0.04940 ETH on the chain, min 0.03402 |
| LetsExchange | `ROBINHOOD` | CASHCAT and ARROW listed, no native ETH. `info` quotes 0.1 ETH into 1154.42 CASHCAT, min 0.06 |

Two providers are mapped without a fillable native-ETH pair at the time of mapping, on the same reasoning: the map is chain-level while fill availability is per-pair and dynamic. `changenow.ts` rebuilds `chainCodeTickerMap` hourly from `exchange/currencies?active=true` filtered to the chain codes the map knows, so an unmapped chain could never appear even once a pair activates. ChangeNow proved the point: the pair it was mapped ahead of fills today with no further change. LetsExchange is the same bet on tokens rather than gas, and costs a declined quote meanwhile, since `checkWhitelistedMainnetCodes` raises a plain `SwapCurrencyError` for a pair the provider cannot serve.

Every provider in the repo was probed against its own live discovery endpoint, not against documentation. ChangeHero, Exolix, Godex and n.exchange carry nothing on this chain and stay `null`. SwapKit's chain code appeared in an earlier sweep but is absent from its keyed provider list. Changelly and Xgram could not be probed: Changelly's API is request-signed, and the available Xgram key returns 401.

Two providers advertise the chain but are deliberately left unmapped: LetsExchange carries only CASHCAT and ARROW there (no native ETH, so it cannot deliver gas), and SwapKit's `HOOD` code turns up in the provider sweep with no tradeable asset behind it.

### Testing

`npm run test` (56 passing), `tsc`, and `verify-repo.sh` all pass.

Exercised end to end on the iOS simulator with this branch served into the app, which is what the mapping is for. An exchange of 529.16 S from My Sonic to a Robinhood Chain wallet quoted 0.006243 ETH "Powered by LI.FI" and executed to the success scene; `eth_getBalance` for the receiving address then returned 6248511112300593 wei and the wallet rendered `0.006248511112300593 ETH`.

That balance unblocked the send that the chain integration could not previously test: 0.001 ETH sent from the funded wallet is transaction [`0x70e4950b5991c9419eff70ea2478c7ae55c5c58adcc3da237d5889394f8fdf62`](https://robinhoodchain.blockscout.com/tx/0x70e4950b5991c9419eff70ea2478c7ae55c5c58adcc3da237d5889394f8fdf62), `success` in block 35308086.

The route was pinned to LI.FI/Rango for the test by disabling competitors locally, because SideShift is geo-blocked from this host's egress; its mapping is verified against its live pair endpoint instead.

One unrelated gap this surfaced: Edge's rates server has no entry for `robinhood`, so ETH on the chain prices at `$0` everywhere, and on the exchange scene that zero destination value trips the "High Price Impact" banner. Cosmetic, fixed where rates are configured rather than here.

Design doc: [`src/docs/robinhood-chain.md`](https://github.com/EdgeApp/edge-currency-accountbased/blob/jon/robinhood-chain/src/docs/robinhood-chain.md), section 7.

Asana: https://app.asana.com/0/1215088146871429/1217382370104887

---

> [!NOTE]
> **Low Risk**
> Additive mapping data verified against live provider APIs; mis-mapping would mainly affect quote availability rather than altering existing swap execution paths.
> 
> **Overview**
> Adds **`robinhood`** as an Edge currency plugin id and wires Robinhood Chain (EVM 4663) into swap routing through **chain-code mappings only**—no quote or engine logic changes. Providers only see a chain when their `src/mappings/*.ts` entry is non-null.
> 
> **Enabled on six partners** with each provider’s own identifier: SideShift (`robinhood`), LI.FI (`out`), Rango (`ROBINHOOD`), ChangeNow (`hood`), Swapuz (`ROBINHOOD`), and LetsExchange (`ROBINHOOD`). Authoritative entries live in `scripts/mappings/*Mappings.ts`; most `src/mappings/*.ts` files are regenerated, with **`changelly`** and **`nexchange`** updated by hand to set `robinhood` → `null` like the other unsupported providers.
> 
> Every other swap partner in the repo gets an explicit **`robinhood` → `null`** mapping so the chain stays invisible where APIs do not support it. **CHANGELOG** records Robinhood Chain swaps via those six integrations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 876a7572a4ad0a17edcced42b62bc87181845e44. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


j0ntz:
## 📸 Test evidence: swap into Robinhood Chain, then send

**lifi quote into robinhood**

lifi quote into robinhood

**swap executed**

swap executed

**hood wallet funded**

hood wallet funded

**send composed**

send composed

**send success**

send success

peachbits (approved):

PR #485, opened 2026-08-14
Add StealthEX swap plugin

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

Adds StealthEX as a centralized swap provider, on the v4 API.

Asana: https://app.asana.com/0/1215088146871429/1217498055202092

**Asset identity.** StealthEX names every asset by a `symbol` plus a `network`, so `src/mappings/stealthex.ts` records both halves per chain: the symbol and network of the chain's native asset, plus the network its tokens live on. A chain's native asset usually sits on `mainnet`, but plenty do not (BNB on `bsc`, AVAX on `avax-c`, WAXP on `wax`, OSMO on `cosmos`, ETH on each L2's own network), which is why the mapping spells out both rather than deriving one from the other.

Token identity comes from the CONTRACT ADDRESS, never from the Edge currency code. The plugin pulls StealthEX's full `/v4/currencies` catalog (paged at 250, about a thousand listings, page 0 then batches of four), indexes it by provider network, and canonicalizes each listing's `contract_address` through that wallet's own `currencyConfig.getTokenId` so the index is keyed by Edge tokenId. Casing differences and junk values in the provider's contract field drop out of that step. A native asset must match the mapped symbol on its network AND carry no contract address, so a token listing can never be resolved for a native swap. A token StealthEX lists without a contract address (USDC on Tron among them) stays unsupported rather than quoted on a ticker match.

The catalog is cached for an hour and replaced atomically, only after every page arrives non-empty. A failed or empty refresh keeps the previous good index and leaves the stamp expired, so the next quote retries instead of caching an outage. The expiry is not theoretical: ARRR and ZANO were both listed on 2026-08-14 and are absent from the catalog on 2026-08-18.

**Rate type.** Quotes use a fixed rate when both assets support one, which is also what makes "quote by the amount I want to receive" work (StealthEX only offers reversed estimation on fixed rates). Both assets advertising fixed support does not guarantee the pair has a fixed route, so a missing route falls back to a floating quote with `isEstimate: true`.

`fetchSwapQuote` runs `checkInvalidTokenIds` first, so the shared same-asset guard and the default invalid-token list apply here as they do in the sibling central plugins.

**Other notes.** Requests go through `io.fetchCors`: StealthEX blocks browser-origin requests, and swap plugins run inside a WebView. Limits come from `POST /v4/rates/range` and map to `SwapBelowLimitError` / `SwapAboveLimitError`; `NoPair` / `NoExchangeRoute` / `RouteIsDisabled` / `MarketUnavailable` map to `SwapCurrencyError` and `NotAllowed` to a reasonless `SwapPermissionError`: StealthEX returns `NotAllowed` for causes it does not name (it was observed refusing a route minutes after the identical call succeeded), so claiming a geographic restriction would be a guess.

Telos and Monad are listed by StealthEX but mapped to `null`: neither listing carries an address format, so which address space they expect cannot be confirmed, and a wrong guess loses funds.

**Errors.** One helper maps a StealthEX error kind to its Edge error for every call site (`/rates/range`, `/rates/estimated-amount`, `/exchanges`), so a kind cannot mean an unsupported pair on one path and a generic failure on another. A catalog outage raises a real provider error rather than `SwapCurrencyError`, so it keeps its error ranking instead of reading as an unsupported pair.

**Fixed orders always carry a locked rate.** A fixed estimate that comes back without `rate.id` is treated as no fixed route: a direct quote falls back to floating with the range, limit check and estimate re-run for that rate, so `isEstimate` always describes the order actually created.

**Max quotes create one order, not two.** `getMaxSwappable` runs the quote function twice, so the quote path is split: `fetchQuote` resolves the route, checks limits and returns both addresses without creating anything; `fetchProbeOrder` builds the probe `spendInfo` from it (the user's own from-address, `skipChecks: true`) so `getMaxSpendable` can price the network fee; and `fetchSwapQuoteInner` is the only caller that POSTs `/exchanges`. Above the maximum the probe quotes the range maximum instead of throwing, because StealthEX rejects an out-of-range amount at `/rates/estimated-amount` with `400 InvalidAmount` and a max swap that fits once the fee is subtracted would otherwise never get priced; the real quote still raises `SwapAboveLimitError`. Measured against the live API, a max quote issued two orders before this and one after.

**Trust boundary.** The deposit amount StealthEX returns becomes a signed spend, so a `from` quote (a max quote arrives here as one) rejects an order whose deposit exceeds the requested amount. A reversed quote pins the receive side, so there is nothing local to bound the source against.

The review rules and the API requirement this integration surfaced are in EdgeApp/edge-exchange-plugins#489: per-endpoint auth, catalog cache expiry, the scope a fixed-then-floating fallback is allowed to cover, and the per-route rate-type reporting StealthEX does not publish.

A compliance and gap report for StealthEX v4 against `docs/API_REQUIREMENTS.md` is published at https://gist.github.com/j0ntz/ae71f231a5939489cf4611325058c00c

Section 7 of that report, account activation, now reads **Pass**, settled with real funded swaps on every reserve network Edge and StealthEX both support: XRP (26.337 TRX for 6.1363 XRP), ALGO (0.0035 ETH on Base for 87.20 ALGO), XLM (1968.5 DGB for 42.697 XLM) and SOL (0.0036226 ETH on Base for 0.09283 SOL). Each destination wallet was created in Edge for the test and did not exist on chain when the swap was placed; each payout created the account at exactly the quoted amount, with nothing needed from Edge or from the user. StealthEX also pre-checks the destination on XRP, where a payout below the 1 XRP base reserve to an unactivated address is refused outright (`NotAllowed`, "This address can not be used") while the same payout to an activated address is accepted. That pre-check is chain-specific: Algorand accepts an order whose payout would land below its 0.1 ALGO minimum balance, and every Stellar route already prices its minimum above the 1 XLM reserve, so the case cannot be ordered there at all. Hedera stays untestable from Edge, because a fresh Edge Hedera wallet has no account id to receive into. The one wart is that the refusal reuses the catch-all `NotAllowed` kind, so this plugin can only surface it as a reasonless `SwapPermissionError`; that is filed as an ask under the report's error-handling section.

Companion GUI PR wires the plugin up: EdgeApp/edge-react-gui#6160

---

> [!NOTE]
> **Medium Risk**
> New swap path moves user funds to provider deposit addresses, but the change is additive and follows established max-quote, limit, and trust-boundary patterns from other central plugins.
> 
> **Overview**
> Adds **StealthEX** as a new centralized swap provider (`stealthex`), wired into the plugin registry and documented in the unreleased CHANGELOG.
> 
> The integration talks to StealthEX **v4** (`/currencies`, `/rates/range`, `/rates/estimated-amount`, `/exchanges`) via `fetchCors` and a Bearer API key. **`src/mappings/stealthex.ts`** maps each Edge chain to StealthEX `symbol` + `network` (and token network), with explicit `null` where support is unclear or absent.
> 
> **Quoting behavior:** the plugin pages and hour-caches the asset catalog, resolves tokens by **canonical contract address** (not ticker), prefers **fixed** rates when the route supports them (required for receive-side / reversed quotes), and falls back to **floating** with `isEstimate: true` when fixed is unavailable. Limits, provider errors, and catalog outages are mapped to the usual Edge swap errors without mislabeling provider failures as unsupported pairs.
> 
> **Max swaps and safety:** `getMaxSwappable` uses a **probe** that reuses `fetchQuote` without POSTing an order (avoiding abandoned deposit addresses); only **`fetchSwapQuoteInner`** creates an exchange. Deposit amounts from the provider are floored to native units, memos are normalized from `extra_id`, and **from** quotes reject deposits above the requested amount at a trust boundary before signing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c0a70ddbb88615bd06797f7cc7953b714545f4c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/mappings/stealthex.ts
+++ b/src/mappings/stealthex.ts
@@ -1,0 +1,315 @@
+/**
+ * StealthEX Exchange Plugin Chain Mapping
+ *
+ * See https://api.stealthex.io/docs/ for the API reference. To list the
+ * supported assets:
+ * `curl -X GET 'https://api.stealthex.io/v4/currencies?limit=250&offset=0' -H 'Authorization: Bearer <your-api-key>' | jq .`
+ *
+ * StealthEX identifies every asset by a `symbol` plus a `network`. A chain's
+ * native asset usually sits on the `mainnet` network, but plenty of chains are
+ * listed under their own network instead (BNB on `bsc`, AVAX on `avax-c`, ETH
+ * on each L2's network), so both halves are spelled out per chain. Tokens use
+ * the chain's `tokenNetwork` together with the token's own symbol.
+ *
+ * This file maps EdgeCurrencyPluginId -> StealthEX chain identity (or null when
+ * StealthEX does not list the chain).
+ */
+
+import { EdgeCurrencyPluginId } from '../util/edgeCurrencyPluginIds'
+
+export interface StealthexChain {
+  /** StealthEX symbol of the chain's native asset */
+  mainnetSymbol: string
+  /** StealthEX network hosting the chain's native asset */
+  mainnetNetwork: string
+  /** StealthEX network hosting the chain's tokens, or null when unsupported */
+  tokenNetwork: string | null
+}
+
+export const stealthex = new Map<EdgeCurrencyPluginId, StealthexChain | null>()
+stealthex.set('abstract', null)
+stealthex.set('algorand', {
+  mainnetSymbol: 'algo',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: 'algo'
+})
+stealthex.set('amoy', null)
+stealthex.set('arbitrum', {
+  mainnetSymbol: 'eth',
+  mainnetNetwork: 'arbitrum',
+  tokenNetwork: 'arbitrum'
+})
+stealthex.set('avalanche', {
+  mainnetSymbol: 'avax',
+  mainnetNetwork: 'avax-c',
+  tokenNetwork: 'avax-c'
+})
+stealthex.set('axelar', {
+  mainnetSymbol: 'axl',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('badcoin', null)
+stealthex.set('base', {
+  mainnetSymbol: 'eth',
+  mainnetNetwork: 'base',
+  tokenNetwork: 'base'
+})
+// StealthEX lists no BEP2 assets:
+stealthex.set('binance', null)
+stealthex.set('binancesmartchain', {
+  mainnetSymbol: 'bnb',
+  mainnetNetwork: 'bsc',
+  tokenNetwork: 'bsc'
+})
+stealthex.set('bitcoin', {
+  mainnetSymbol: 'btc',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('bitcoincash', {
+  mainnetSymbol: 'bch',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('bitcoincashtestnet', null)
+stealthex.set('bitcoingold', null)
+stealthex.set('bitcoingoldtestnet', null)
+stealthex.set('bitcoinsv', {
+  mainnetSymbol: 'bsv',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('bitcointestnet', null)
+stealthex.set('bitcointestnet4', null)
+stealthex.set('bobevm', null)
+stealthex.set('botanix', null)
+stealthex.set('calibration', null)
+stealthex.set('cardano', {
+  mainnetSymbol: 'ada',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: 'ada'
+})
+stealthex.set('cardanotestnet', null)
+stealthex.set('celo', {
+  mainnetSymbol: 'celo',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: 'celo'
+})
+stealthex.set('coreum', null)
+stealthex.set('cosmoshub', {
+  mainnetSymbol: 'atom',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('dash', {
+  mainnetSymbol: 'dash',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('digibyte', {
+  mainnetSymbol: 'dgb',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('dogecoin', {
+  mainnetSymbol: 'doge',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('eboost', null)
+stealthex.set('ecash', {
+  mainnetSymbol: 'xec',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+// EOS was rebranded to Vaulta, and StealthEX lists it under its new `a` symbol:
+stealthex.set('eos', {
+  mainnetSymbol: 'a',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('ethDev', null)
+stealthex.set('ethereum', {
+  mainnetSymbol: 'eth',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: 'eth'
+})
+stealthex.set('ethereumclassic', {
+  mainnetSymbol: 'etc',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('ethereumpow', {
+  mainnetSymbol: 'ethw',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+// StealthEX delisted FTM when Fantom became Sonic:
+stealthex.set('fantom', null)
+stealthex.set('feathercoin', null)
+stealthex.set('filecoin', {
+  mainnetSymbol: 'fil',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+// StealthEX only supports Filecoin's native address space:
+stealthex.set('filecoinfevm', null)
+stealthex.set('filecoinfevmcalibration', null)
+stealthex.set('fio', {
+  mainnetSymbol: 'fio',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('groestlcoin', null)
+stealthex.set('hedera', {
+  mainnetSymbol: 'hbar',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('holesky', null)
+stealthex.set('hyperevm', {
+  mainnetSymbol: 'hype',
+  mainnetNetwork: 'hyperevm',
+  tokenNetwork: 'hyperevm'
+})
+stealthex.set('liberland', null)
+stealthex.set('liberlandtestnet', null)
+stealthex.set('litecoin', {
+  mainnetSymbol: 'ltc',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('mayachain', null)
+// StealthEX's `monad` listing carries no address format, so the address space it
+// expects cannot be confirmed:
+stealthex.set('monad', null)
+stealthex.set('monero', {
+  mainnetSymbol: 'xmr',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+// StealthEX only lists NYM as an Ethereum token, not the Nyx chain:
+stealthex.set('nym', null)
+stealthex.set('opbnb', null)
+stealthex.set('optimism', {
+  mainnetSymbol: 'eth',
+  mainnetNetwork: 'optimism',
+  tokenNetwork: 'optimism'
+})
+stealthex.set('osmosis', {
+  mainnetSymbol: 'osmo',
+  mainnetNetwork: 'cosmos',
+  tokenNetwork: null
+})
+stealthex.set('piratechain', {
+  mainnetSymbol: 'arrr',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('pivx', {
+  mainnetSymbol: 'pivx',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('polkadot', {
+  mainnetSymbol: 'dot',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('polygon', {
+  mainnetSymbol: 'pol',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: 'matic'
+})
+stealthex.set('pulsechain', {
+  mainnetSymbol: 'pls',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('qtum', {
+  mainnetSymbol: 'qtum',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('ravencoin', null)
+stealthex.set('ripple', {
+  mainnetSymbol: 'xrp',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: 'xrp'
+})
+stealthex.set('rsk', null)
+stealthex.set('sepolia', null)
+stealthex.set('smartcash', null)
+stealthex.set('solana', {
+  mainnetSymbol: 'sol',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: 'sol'
+})
+stealthex.set('sonic', {
+  mainnetSymbol: 's',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('stellar', {
+  mainnetSymbol: 'xlm',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: 'xlm'
+})
+stealthex.set('sui', {
+  mainnetSymbol: 'sui',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: 'sui'
+})
+stealthex.set('suitestnet', null)
+// StealthEX's `tlos` listing carries no address format, so whether it expects
+// native Telos or Telos EVM addresses cannot be confirmed:
+stealthex.set('telos', null)
+stealthex.set('tezos', {
+  mainnetSymbol: 'xtz',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: 'xtz'
+})
+stealthex.set('thorchainrune', {
+  mainnetSymbol: 'rune',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('thorchainrunestagenet', null)
+stealthex.set('ton', {
+  mainnetSymbol: 'ton',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: 'ton'
+})
+stealthex.set('tron', {
+  mainnetSymbol: 'trx',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: 'trx'
+})
+stealthex.set('ufo', null)
+stealthex.set('vertcoin', null)
+stealthex.set('wax', {
+  mainnetSymbol: 'waxp',
+  mainnetNetwork: 'wax',
+  tokenNetwork: null
+})
+stealthex.set('zano', {
+  mainnetSymbol: 'zano',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('zcash', {
+  mainnetSymbol: 'zec',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('zcoin', {
+  mainnetSymbol: 'firo',
+  mainnetNetwork: 'mainnet',
+  tokenNetwork: null
+})
+stealthex.set('zksync', {
+  mainnetSymbol: 'eth',
+  mainnetNetwork: 'zksync',
+  tokenNetwork: 'zksync'
+})
--- a/src/swap/central/stealthex.ts
+++ b/src/swap/central/stealthex.ts
@@ -1,0 +1,724 @@
+import { ceil, floor, gt, lt } from 'biggystring'
+import {
+  asArray,
+  asDate,
+  asEither,
+  asJSON,
+  asMaybe,
+  asNull,
+  asNumber,
+  asObject,
+  asOptional,
+  asString
+} from 'cleaners'
+import {
+  EdgeCorePluginOptions,
+  EdgeCurrencyWallet,
+  EdgeMemo,
+  EdgeSpendInfo,
+  EdgeSwapInfo,
+  EdgeSwapPlugin,
+  EdgeSwapQuote,
+  EdgeSwapRequest,
+  EdgeToken,
+  EdgeTokenId,
+  SwapAboveLimitError,
+  SwapBelowLimitError,
+  SwapCurrencyError,
+  SwapPermissionError
+} from 'edge-core-js/types'
+
+import {
+  stealthex as stealthexMapping,
+  StealthexChain
+} from '../../mappings/stealthex'
+import { EdgeCurrencyPluginId } from '../../util/edgeCurrencyPluginIds'
+import {
+  checkInvalidTokenIds,
+  ensureInFuture,
+  getMaxSwappable,
+  InvalidTokenIds,
+  makeSwapPluginQuote,
+  SwapOrder
+} from '../../util/swapHelpers'
+import {
+  convertRequest,
+  denominationToNative,
+  getAddress,
+  memoType,
+  nativeToDenomination
+} from '../../util/utils'
+import { asNumberString, EdgeSwapRequestPlugin, StringMap } from '../types'
+
+const pluginId = 'stealthex'
+
+export const swapInfo: EdgeSwapInfo = {
+  pluginId,
+  isDex: false,
+  displayName: 'StealthEX',
+  supportEmail: 'support@stealthex.io'
+}
+
+const asInitOptions = asObject({
+  apiKey: asString
+})
+
+const API_BASE_URL = 'https://api.stealthex.io/v4'
+const ORDER_BASE_URL = 'https://stealthex.io/exchange/?id='
+
+const addressTypeMap: StringMap = {
+  zcash: 'transparentAddress'
+}
+
+/** StealthEX identifies an asset by its symbol on a specific network */
+interface StealthexAsset {
+  symbol: string
+  network: string
+}
+
+type StealthexRate = 'fixed' | 'floating'
+
+/**
+ * `reversed` estimation quotes by the amount the user receives, and StealthEX
+ * only offers it on fixed rate routes.
+ */
+type StealthexEstimation = 'direct' | 'reversed'
+
+/**
+ * One asset in StealthEX's catalog. `rates` is cleaned as plain strings so a
+ * rate type StealthEX adds later cannot invalidate every listing.
+ */
+const asStealthexCurrency = asObject({
+  symbol: asString,
+  network: asString,
+  rates: asArray(asString),
+  contract_address: asEither(asString, asNull)
+})
+type StealthexCurrency = ReturnType<typeof asStealthexCurrency>
+
+/** A malformed listing drops itself rather than the whole catalog page */
+const asStealthexCurrencies = asArray(asMaybe(asStealthexCurrency))
+
+/** The body both `/rates/estimated-amount` and `/exchanges` are built on */
+interface StealthexEstimateBody {
+  route: { from: StealthexAsset; to: StealthexAsset }
+  estimation: StealthexEstimation
+  rate: StealthexRate
+  amount: number
+}
+
+const asStealthexRange = asObject({
+  min_amount: asNumberString,
+  max_amount: asOptional(asNumberString)
+})
+
+const asStealthexEstimate = asObject({
+  estimated_amount: asNumberString,
+  // Only fixed rate estimates come with a rate to lock in:
+  rate: asOptional(asObject({ id: asString }))
+})
+
+/**
+ * A deposit memo / destination tag. StealthEX documents it as a string, but a
+ * numeric tag (XRP, and other tag chains) would arrive as a JSON number, and a
+ * cleaner that only accepts strings would fail the whole order response. An
+ * absent or blank value means the chain takes no memo.
+ */
+const asStealthexExtraId = asOptional(asEither(asString, asNumber, asNull))
+
+const asStealthexExchange = asObject({
+  id: asString,
+  deposit: asObject({
+    expected_amount: asNumberString,
+    address: asString,
+    extra_id: asStealthexExtraId
+  }),
+  withdrawal: asObject({
+    expected_amount: asNumberString
+  }),
+  created_at: asDate,
+  expires_at: asOptional(asDate)
+})
+
+/**
+ * Every StealthEX failure comes back as `{err: {kind, details}}`, and the kind
+ * is what distinguishes a bad pair from a bad amount.
+ */
+const asStealthexError = asMaybe(
+  asJSON(asObject({ err: asObject({ kind: asString, details: asString }) }))
+)
+
+/** Error kinds that mean StealthEX cannot swap this pair at all */
+const NO_ROUTE_KINDS = [
+  'NoPair',
+  'NoExchangeRoute',
+  'RouteIsDisabled',
+  'MarketUnavailable'
+]
+
+const INVALID_TOKEN_IDS: InvalidTokenIds = {
+  from: {},
+  to: {}
+}
+
+/** Fixed rate orders expire on their own, so a floating order needs a window */
+const FLOATING_EXPIRATION_MS = 1000 * 60 * 20
+
+/** `/currencies` pages at 250 listings, and the catalog runs about 1000 */
+const CATALOG_PAGE_SIZE = 250
+const CATALOG_MAX_PAGES = 40
+/** Pages fetched at once after the first, so a quote is not stuck serially */
+const CATALOG_PAGE_BATCH = 4
+const CATALOG_CACHE_MS = 1000 * 60 * 60
+
+export function makeStealthexPlugin(
+  opts: EdgeCorePluginOptions
+): EdgeSwapPlugin {
+  const { io, log } = opts
+  // StealthEX blocks browser-origin requests via CORS, and swap plugins run
+  // inside a WebView:
+  const { fetchCors = io.fetch } = io
+  const { apiKey } = asInitOptions(opts.initOptions)
+
+  const headers = {
+    'Content-Type': 'application/json',
+    Accept: 'application/json',
+    Authorization: `Bearer ${apiKey}`
+  }
+
+  /**
+   * StealthEX's whole asset catalog, indexed by provider network. It carries
+   * each listing's symbol, contract address and supported rate types, which is
+   * everything a quote needs to identify an asset.
+   */
+  let catalog = new Map<string, StealthexCurrency[]>()
+  let catalogUpdated = 0
+  let catalogErrorKind: string | undefined
+  let catalogFetch: Promise<void> | undefined
+
+  /**
+   * Edge tokenId to catalog listing, per `pluginId:network`, derived from
+   * `catalog`. Canonicalizing a contract address costs a call into the currency
+   * plugin and one network carries hundreds of listings, so a derived index
+   * lives until the catalog it came from is replaced.
+   */
+  const tokenIndexes = new Map<
+    string,
+    { stamp: number; byTokenId: Map<string, StealthexCurrency> }
+  >()
+
+  /**
+   * Returns the parsed JSON body, or the StealthEX error kind when the request
+   * failed with a recognizable error envelope.
+   */
+  const fetchStealthex = async (
+    path: string,
+    body?: unknown
+  ): Promise<{ json?: unknown; errorKind?: string }> => {
+    const response = await fetchCors(`${API_BASE_URL}${path}`, {
+      headers,
+      method: body == null ? 'GET' : 'POST',
+      body: body == null ? undefined : JSON.stringify(body)
+    })
+    const text = await response.text()
+
+    if (!response.ok) {
+      log.warn(`StealthEX ${path} returned ${response.status}: ${text}`)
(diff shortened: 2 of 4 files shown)


j0ntz:
## 📸 Test evidence

**stealthex quote**

stealthex quote

**stealthex success**

stealthex success

**stealthex tx details**

stealthex tx details

src/swap/central/stealthex.ts
```diff
@@ -0,0 +1,499 @@
...
+   * could send funds to the wrong token's deposit address.
+   */
+  const checkContractAddress = (
+    currency: StealthexCurrency,
+    contractAddress: string | undefined,
+    request: EdgeSwapRequestPlugin
+  ): void => {
+    // StealthEX leaves the contract address off some token listings, which
+    // leaves nothing to compare against:
+    if (currency.contract_address == null) return
```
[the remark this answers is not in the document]
j0ntz: Valid, and fixed in 5a63ddd.

`checkContractAddress` now branches on the Edge side rather than the StealthEX side. A native asset (no Edge contract address) requires StealthEX to have no contract address either, so a token listing can never be resolved for a native swap. A token requires a present, matching contract address; a listing with no contract address is now rejected instead of accepted on a symbol match.

That drops the tokens StealthEX lists without a contract address, USDC on Tron among them, which is the right trade: without the contract address there is nothing that establishes identity, and a colliding ticker would point the deposit at the wrong asset.

Verified against the live API through the plugin itself: USDC on Ethereum (contract present and matching) still quotes, USDC on Tron (contract absent) now throws `SwapCurrencyError`.

src/swap/central/stealthex.ts
```diff
@@ -0,0 +1,508 @@
...
+
+  const out: EdgeSwapPlugin = {
+    swapInfo,
+
+    async fetchSwapQuote(req: EdgeSwapRequest): Promise<EdgeSwapQuote> {
+      const request = convertRequest(req)
+      const newRequest = await getMaxSwappable(getQuote, request)
+      const swapOrder = await getQuote(newRequest)
+      return await makeSwapPluginQuote(swapOrder)
+    }
```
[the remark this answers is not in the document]
j0ntz: Valid, and fixed in 8ae9a4d. `fetchSwapQuote` now calls `checkInvalidTokenIds(INVALID_TOKEN_IDS, request, swapInfo)` with an empty plugin-specific map, matching the sibling central plugins, so the shared `isSameAsset` guard and the default invalid-token list both apply.

Verified against the live API through the plugin: a BTC to BTC request now throws `SwapCurrencyError` before any StealthEX call, where it previously reached the range request.

src/swap/central/stealthex.ts
```diff
@@ -0,0 +1,518 @@
...
+  network: asString,
+  rates: asArray(asStealthexRate),
+  contract_address: asEither(asString, asNull)
+})
+type StealthexCurrency = ReturnType<typeof asStealthexCurrency>
+
+const asStealthexRange = asObject({
+  min_amount: asNumberString,
+  max_amount: asOptional(asNumberString)
+})
```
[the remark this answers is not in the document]
j0ntz: Not a defect. `asOptional` in cleaners tests `raw == null`, which covers JSON `null` as well as a missing key, so a `null` value cleans to `undefined` rather than throwing:

```
asObject({ max_amount: asOptional(asNumberString) })({ min_amount: 0.1, max_amount: null })
// => { min_amount: '0.1' }   (no throw)
asObject({ expires_at: asOptional(asDate) })({ expires_at: null })
// => { }                     (no throw)
```

The live floating path already exercises exactly this shape. A BTC to ARRR quote (no fixed route, so it falls back to floating) returns `{"min_amount":0.0004774713119320081,"max_amount":null}` from `/v4/rates/range` and `"expires_at": null` from `/v4/exchanges`, and it produces a valid quote with the synthetic 20 minute window and `isEstimate: true`.

Resolving as a false positive.

src/swap/central/stealthex.ts
```diff
@@ -0,0 +1,518 @@
...
+  deposit: asObject({
+    expected_amount: asNumberString,
+    address: asString,
+    extra_id: asEither(asString, asNull)
+  }),
+  withdrawal: asObject({
+    expected_amount: asNumberString
+  }),
+  created_at: asDate,
+  expires_at: asOptional(asDate)
```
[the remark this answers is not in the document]
j0ntz: Not a defect. `asOptional` in cleaners tests `raw == null`, which covers JSON `null` as well as a missing key, so a `null` value cleans to `undefined` rather than throwing:

```
asObject({ max_amount: asOptional(asNumberString) })({ min_amount: 0.1, max_amount: null })
// => { min_amount: '0.1' }   (no throw)
asObject({ expires_at: asOptional(asDate) })({ expires_at: null })
// => { }                     (no throw)
```

The live floating path already exercises exactly this shape. A BTC to ARRR quote (no fixed route, so it falls back to floating) returns `{"min_amount":0.0004774713119320081,"max_amount":null}` from `/v4/rates/range` and `"expires_at": null` from `/v4/exchanges`, and it produces a valid quote with the synthetic 20 minute window and `isEstimate: true`.

Resolving as a false positive.

src/swap/central/stealthex.ts
```diff
@@ -0,0 +1,518 @@
...
+        symbol: chain.mainnetSymbol,
+        network: chain.mainnetNetwork
+      }
+    }
+    if (chain.tokenNetwork == null) return undefined
+    return {
+      symbol: currencyCode.toLowerCase(),
+      network: chain.tokenNetwork
+    }
+  }
```
paullinator: **Issue 1 - Must fix: resolve tokens from the contract address, not from the Edge currency code**

For a token this builds the StealthEX asset as `currencyCode.toLowerCase()` plus a statically mapped `tokenNetwork`, then `checkContractAddress` verifies afterward that the guessed listing happens to carry the expected contract. Two problems follow from that ordering:

- Ticker symbols are not unique and Edge does not always use the same symbol StealthEX does, so a supported token gets rejected whenever the two spellings differ.
- The contract address Edge already has is only ever used to reject a guess. It is never used to discover which StealthEX listing is the correct one.

The contract address should be the input to asset resolution rather than a post-hoc check. StealthEX's paginated currency catalog (`GET /v4/currencies`) returns `symbol`, `network`, and `contract_address` together, which is enough to build that lookup.

Suggested implementation, which follows the workflow other providers here use:

1. Fetch the currency catalog across all pages, keeping each listing's `symbol`, `network`, and `contract_address`.
2. Validate the response with cleaners and build the index from the cleaned data.
3. Canonicalize each provider contract through the relevant wallet's `currencyConfig.getTokenId` rather than comparing raw contract strings, so provider casing and formatting differences cannot cause a false mismatch.
4. Key the index by provider network plus canonical Edge token ID, and have it yield the exact `{ symbol, network }` pair the quote and order requests need.
5. Resolve both the source and destination assets through that index before calling `/rates/range`, `/rates/estimated-amount`, or `/exchanges`. If either side is missing from the index, throw `SwapCurrencyError`; never fall back to guessing a symbol from the currency code.
6. Cache the index with a bounded TTL (an hour is typical). On refresh, replace the cache atomically and only after a successful non-empty fetch. Keep the previous good index on a transient failure, and leave the timestamp expired so the next quote retries rather than caching an outage for the full TTL.

That keeps the contract address Edge supplies as the source of truth while still sending StealthEX the symbol/network identifiers its API requires.

src/mappings/stealthex.ts
```diff
@@ -0,0 +1,315 @@
...
+import { EdgeCurrencyPluginId } from '../util/edgeCurrencyPluginIds'
+
+export interface StealthexChain {
+  /** StealthEX symbol of the chain's native asset */
+  mainnetSymbol: string
+  /** StealthEX network hosting the chain's native asset */
+  mainnetNetwork: string
+  /** StealthEX network hosting the chain's tokens, or null when unsupported */
+  tokenNetwork: string | null
+}
```
paullinator, line 27: **Issue 2 - Warning: EVM chains are identified by static network names instead of the numeric chain ID**

Every chain, including each EVM chain, is pinned to a hardcoded provider network string (`mainnet`, `bsc`, `arbitrum`, `avax-c`, and so on), and the plugin never reads `currencyInfo.evmChainId`. As a result, each newly supported EVM chain requires editing this table and shipping a plugin release, even though the wallet already knows both the chain ID and the token contract.

The goal is for new EVM chains to work with no plugin change. That requires StealthEX to expose the standard numeric EVM chain ID in its currency catalog and to accept it when identifying assets for quotes and orders. Worth raising with them.

Once the API supports it, EVM assets should be resolved by chain ID plus canonical token contract through the index described in issue 1, leaving this static table to cover non-EVM chains only.

src/swap/central/stealthex.ts
```diff
@@ -0,0 +1,518 @@
...
+    Authorization: `Bearer ${apiKey}`
+  }
+
+  /**
+   * Resolved currencies keyed by `symbol:network`. Contract addresses and
+   * supported rate types do not change, so successful lookups live for the
+   * lifetime of the plugin. Failures are never cached, so a newly listed asset
+   * works without restarting the app.
+   */
+  const currencyCache = new Map<string, StealthexCurrency>()
```
paullinator: **Issue 3 - Must fix: the currency cache never expires**

The comment states that contract addresses and rate types do not change, so successful lookups live for the lifetime of the plugin. That assumption does not hold in practice: providers rename listings, delist assets, correct a wrong contract address, and add or remove fixed-rate support. Any of those leaves this cache serving stale data for as long as the app stays open, and the app can stay open for days.

Replace this with the bounded-TTL catalog cache described in issue 1: refresh after expiry, swap in a new index only after a successful non-empty fetch, and keep serving the previous good index if a refresh fails. Do not stamp a failed or empty refresh as a successful one, or an outage gets cached for the full TTL and every token quote fails as an unsupported pair until it lapses.

src/swap/central/stealthex.ts
```diff
@@ -0,0 +1,518 @@
...
+  ): Promise<StealthexCurrency> => {
+    const cacheKey = `${asset.symbol}:${asset.network}`
+    const cached = currencyCache.get(cacheKey)
+    if (cached != null) return cached
+
+    const path = `/currencies/${asset.symbol}/${asset.network}`
+    const { json, errorKind } = await fetchStealthex(path)
+    if (errorKind != null || json == null) {
+      throw new SwapCurrencyError(swapInfo, request)
+    }
```
paullinator: **Issue 4 - Must fix: every currency lookup failure is reported as an unsupported pair**

This collapses all failures into `SwapCurrencyError`: a genuinely unknown asset, but also an auth failure, a region refusal, a rate limit, a 5xx, and any unrecognized error envelope. The estimate and exchange paths get this right, mapping `NotAllowed` to `SwapPermissionError` and anything unexpected to `unexpectedError`. Currency lookup should use the same policy.

Because `SwapCurrencyError` ranks low, an outage here does not surface as a provider problem. It quietly drops StealthEX from the quote race and the user is told the pair is unsupported, which also makes the failure hard to spot in logs.

Map `NotAllowed` to `SwapPermissionError`, map genuine missing-asset responses to `SwapCurrencyError`, and throw `unexpectedError(path, errorKind)` for everything else.

src/swap/central/stealthex.ts
```diff
@@ -0,0 +1,518 @@
...
+      const { json, errorKind } = await fetchStealthex('/rates/range', {
+        route,
+        estimation,
+        rate
+      })
+      if (errorKind != null || json == null) {
+        if (errorKind != null && !NO_ROUTE_KINDS.includes(errorKind)) {
+          throw unexpectedError('/rates/range', errorKind)
+        }
+        return undefined
```
paullinator: **Issue 5 - Must fix: `/rates/range` maps `NotAllowed` inconsistently**

Here any kind outside `NO_ROUTE_KINDS` becomes `unexpectedError`, including `NotAllowed`. The estimate and exchange paths treat that same kind as `SwapPermissionError`. So whether a region or permission refusal reaches the user as a typed permission error or as a generic plugin failure depends only on which call happened to hit it first, and the range call is usually first.

Apply one shared error-kind mapping at every stage. Given this same mapping now appears in three places with only the path differing, it is worth extracting a small helper that takes the path and error kind and throws the right Edge error, so the three call sites cannot drift apart again.

src/swap/central/stealthex.ts
```diff
@@ -0,0 +1,518 @@
...
+      getAddress(fromWallet, addressTypeMap[fromWallet.currencyInfo.pluginId]),
+      getAddress(toWallet, addressTypeMap[toWallet.currencyInfo.pluginId])
+    ])
+
+    const exchangeReply = await fetchStealthex('/exchanges', {
+      ...estimateBody,
+      rate_id: estimate.rate?.id,
+      address: toAddress,
+      refund_address: fromAddress
+    })
```
paullinator, line 754: **Issue 6 - Must fix: fixed-rate orders are created without requiring `rate_id`**

`rate_id: estimate.rate?.id` is optional here, but `isEstimate` is set from the local `rate === 'fixed'` flag rather than from what was actually sent. When the estimate comes back without `rate.id`, the order is created with no locked rate while the saved action still claims the quote is not an estimate. The user is then shown a guaranteed amount the order never locked, and the payout can differ from the quote.

When `rate === 'fixed'`, require `estimate.rate?.id` before calling `/exchanges` and send it as a required field. If it is absent, either fail the quote or explicitly fall back to floating and set `isEstimate: true`, so `isEstimate` always reflects the order that was actually created.

paullinator (changes requested):
Review of the StealthEX swap plugin. Six items, numbered for reference: 1 (must fix), 2 (warning), 3-6 (must fix). Item 1 is the main one and includes a suggested implementation approach.

1. **Must fix** - token identity is guessed from the Edge currency code instead of resolved from the contract address.
2. **Warning** - EVM chains are selected through a static network table rather than the numeric chain ID.
3. **Must fix** - resolved currencies are cached for the lifetime of the plugin with no expiry.
4. **Must fix** - every currency lookup failure is reported as an unsupported pair.
5. **Must fix** - `/rates/range` maps `NotAllowed` differently from the estimate and exchange paths.
6. **Must fix** - fixed-rate orders are created without requiring `rate_id`.

src/swap/central/stealthex.ts
j0ntz: Reworked to resolve assets from the catalog instead of from the currency code. `/v4/currencies` is now fetched in full (5 pages of 250, about 1020 listings), cleaned per listing so one malformed entry cannot drop a page, and indexed by provider network; a token side resolves by canonicalizing each listing's `contract_address` through that wallet's `currencyConfig.getTokenId` and matching the request's tokenId, and a native side requires the mapped symbol on its network with no contract address of its own. A miss throws `SwapCurrencyError` with no ticker fallback, and both `checkContractAddress` and the symbol guess are gone. Driven through the plugin against the live API: BTC to USDC on Ethereum (order `w8GA1dqxc`) and BTC to USDT on Tron (order `f6Ug6OAH8`) quote and create fixed orders, while USDC on Tron stays refused because its listing carries `contract_address: null`.

src/mappings/stealthex.ts
j0ntz: Confirmed against the live API: `GET /v4/currencies/networks` returns bare network names and the `Currency` schema carries no chain-id field, so resolving EVM assets by numeric chain id needs a StealthEX API change. Passing `contract_address` inside a quote `route` is accepted and silently ignored (byte-identical 200 to the same request without it), so today an asset is identified by symbol plus network only. Written up with the other API gaps in the compliance report at https://gist.github.com/j0ntz/ae71f231a5939489cf4611325058c00c ; the static table stays until StealthEX publishes the chain id, at which point EVM assets move onto the contract index from issue 1.

src/swap/central/stealthex.ts
j0ntz: Replaced with a bounded-TTL catalog cache: one hour, refreshed on the first quote after expiry, and the new index is installed only after every page arrived and the result is non-empty. A failed or empty refresh keeps the previous good index and leaves the stamp expired, so the next quote retries instead of caching an outage for the hour. The staleness is not hypothetical: ARRR and ZANO were both listed on 2026-08-14 and are absent from the catalog today, so the old lifetime cache would have kept quoting delisted assets for as long as the app stayed open.

src/swap/central/stealthex.ts
j0ntz: Currency resolution no longer collapses every failure into `SwapCurrencyError`. A catalog refresh that fails or comes back empty is logged and leaves the previous good index serving; when there is no usable index at all the quote throws a real provider error, with `NotAllowed` raised as `SwapPermissionError`, so an outage surfaces as a provider failure and keeps its error ranking instead of reading as an unsupported pair. `SwapCurrencyError` is now raised only for an asset genuinely absent from a good catalog, an unmapped chain, or a chain StealthEX lists with no token network.

src/swap/central/stealthex.ts
j0ntz: Extracted `stealthexError(path, errorKind, request)` and routed `/rates/range`, `/rates/estimated-amount` and `/exchanges` through it, so a kind now maps identically wherever it lands: no-route kinds to `SwapCurrencyError`, `NotAllowed` to `SwapPermissionError`, anything else to the logged unexpected error carrying the path. The range path keeps exactly one local behavior, treating a no-route kind as 'this rate type has no route' so the caller can fall back to the other one, and every other kind goes through the shared mapping.

src/swap/central/stealthex.ts
j0ntz: A fixed estimate that returns without `rate.id` is now treated as no fixed route: it warns and the attempt returns empty, and a direct quote then falls back to floating with the range, limit check and estimate re-run for that rate, so `isEstimate` always describes the order actually created. `rate_id` is sent whenever the estimate carried one, which after that guard is every fixed order. A reversed quote has no floating equivalent (StealthEX rejects reversed floating estimates outright), so it fails with `SwapCurrencyError` rather than creating an unlocked order behind a guaranteed-looking quote.

j0ntz:
### Repo docs added on this branch

`docs: add AGENTS.md and .cursor/BUGBOT.md` distills the findings that keep recurring on plugin PRs into two files the agents read automatically.

**`AGENTS.md`** (repo root): what the repo is, the command table (install/prepare, `npm run verify`, `npm run mapctl`), a docs index, and the integration invariants that fail silently or only under production traffic:

- new central plugins start from `src/swap/central/template.ts`, and `docs/API_REQUIREMENTS.md` is read before writing code
- `fetchSwapQuote` calls `checkInvalidTokenIds` even with an empty plugin-specific map
- token identity resolves by CONTRACT ADDRESS, never by Edge currency code, and a native asset requires the provider listing to carry no contract address either
- every `denominationToNative` result is rounded to integer atomic units before it reaches an order, limit or quote
- provider failures map to the typed swap errors precisely, and the same provider error kind maps identically across the estimate, exchange and range paths
- an empty-string memo or extra id from a provider is absent, never an `EdgeMemo`
- credentials go per endpoint, not globally
- provider catalog caches need an expiry or a re-fetch-on-miss path

**`.cursor/BUGBOT.md`**: the same invariants written as review rules (`shared-guards-required`, `native-amount-rounding`, `error-classification`, `per-endpoint-auth`, `empty-memo-absent`, `catalog-cache-expiry`, `typed-catch`), plus two known-good patterns the bot should not flag: `asOptional` accepting JSON `null`, and the template's raw error logging in `fetchSwapQuoteInner`.

Issues 1, 3, 4, 5 and 6 of this review are those invariants, and this round's fixup brings the plugin in line with each.

### StealthEX API compliance and gap report

Issue 2 (numeric EVM chain id) cannot be fixed on this side, so it is written up rather than coded, together with every other gap against `docs/API_REQUIREMENTS.md` at tip:

https://gist.github.com/j0ntz/ae71f231a5939489cf4611325058c00c

The load-bearing gaps, each verified against the live API on 2026-08-18:

- **Asset identification (section 1).** Quote and order requests identify an asset by `symbol` plus a provider network name and nothing else. A `contract_address` placed inside a `route` is accepted and silently ignored (byte-identical 200 to the request without it), and no numeric chain id exists in `/v4/currencies` or `/v4/currencies/networks`. That is why the plugin keeps a static chain table and builds its own contract index client-side.
- **Transaction status (section 5).** `GET /v4/exchanges/{id}` returns 401 without the API key; the spec requires an unauthenticated status endpoint.
- **Error handling (section 3).** One error per response, and both below-minimum and above-maximum come back as `InvalidAmount` / "Amount is out of range" with no amounts and no direction, so the plugin calls `/rates/range` before every estimate just to recover the limit.
- **Reporting (section 6).** Pagination and date filtering are there, but records carry no terminal-status timestamp and no token or chain identifiers for either leg.

Order identification and the unauthenticated status page pass as-is.

j0ntz:
All six items handled. 1: assets now resolve from the full `/v4/currencies` catalog, indexed by provider network and keyed by the Edge tokenId that each listing's `contract_address` canonicalizes to through `currencyConfig.getTokenId`; the currency code no longer participates in token identity. 3: the catalog cache is bounded at one hour, replaced atomically only after a complete non-empty fetch, with the previous index kept and the stamp left expired on a failure. 4: currency resolution no longer collapses into `SwapCurrencyError`; a catalog outage raises a real provider error and `NotAllowed` becomes `SwapPermissionError`. 5: one `stealthexError(path, kind, request)` helper now serves `/rates/range`, `/rates/estimated-amount` and `/exchanges`. 6: a fixed estimate without `rate.id` is treated as no fixed route, so a direct quote falls back to floating and `isEstimate` always describes the order created. 2 needs a StealthEX API change (no numeric chain id exists in their catalog, and a `contract_address` inside a quote `route` is accepted and ignored); it is documented with the other gaps at https://gist.github.com/j0ntz/ae71f231a5939489cf4611325058c00c and summarized in the PR comment above.

src/swap/central/stealthex.ts
```diff
@@ -0,0 +1,674 @@
...
+  // Only fixed rate estimates come with a rate to lock in:
+  rate: asOptional(asObject({ id: asString }))
+})
+
+const asStealthexExchange = asObject({
+  id: asString,
+  deposit: asObject({
+    expected_amount: asNumberString,
+    address: asString,
+    extra_id: asEither(asString, asNull)
```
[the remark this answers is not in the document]
j0ntz: Fixed. `extra_id` now cleans through `asOptional(asEither(asString, asNumber, asNull))`, so a numeric destination tag, a null, and an absent key all clean without throwing, and the memo is built from `String(extraId)` with an empty string still treated as absent (no `EdgeMemo` emitted). StealthEX's own schema types the field as string-or-null, but a tag chain answering with a JSON number would otherwise have failed the entire order response after the deposit address was already issued.

src/swap/central/stealthex.ts
```diff
@@ -0,0 +1,674 @@
...
+    const limitDirection = quoteFor === 'to' ? 'to' : 'from'
+
+    const checkLimits = (range: ReturnType<typeof asStealthexRange>): void => {
+      if (lt(quoteAmount, range.min_amount)) {
+        throw new SwapBelowLimitError(
+          swapInfo,
+          floor(
+            denominationToNative(quoteWallet, range.min_amount, quoteTokenId),
+            0
+          ),
```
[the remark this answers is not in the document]
j0ntz: Fixed. Bounds now round INWARD: the minimum uses `ceil` and the maximum keeps `floor`, so a rounded limit can never fall outside what StealthEX accepts and hand the user an amount that fails again.

src/swap/central/stealthex.ts
```diff
@@ -0,0 +1,674 @@
...
+      }
+      return { rate, estimate, body }
+    }
+
+    // Both assets supporting fixed rates does not guarantee the pair has a
+    // fixed rate route, so fall back to floating when the fixed one is missing.
+    // A reversed estimate has no floating equivalent, so it never falls back.
+    const attempt =
+      (fixedSupported ? await attemptRate('fixed') : undefined) ??
+      (estimation === 'direct' ? await attemptRate('floating') : undefined)
```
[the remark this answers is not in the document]
j0ntz: Fixed. `checkLimits` now RETURNS the limit error instead of throwing; `attemptRate` records it and returns empty, so the fixed-then-floating chain still tries floating when only the fixed range rejects the amount, and the recorded error is thrown only if no rate type works. Confirmed against the live API through the plugin: 1000 BTC to ETH exceeds the fixed maximum of 220 and now quotes on the floating route instead of failing, and a below-minimum amount reports the floating minimum, which is the smallest amount that actually works.

src/swap/central/stealthex.ts
```diff
@@ -0,0 +1,724 @@
...
+        const tokenId = await wallet.currencyConfig.getTokenId(token)
+        if (tokenId != null) byTokenId.set(tokenId, currency)
+      } catch (error: unknown) {
+        // Not an address this chain recognizes, so not this chain's token
+      }
+    }
+
+    tokenIndexes.set(key, { stamp: catalogUpdated, byTokenId })
+    return byTokenId
+  }
```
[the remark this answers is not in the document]
j0ntz: Fixed. `getTokenIndex` now snapshots both the catalog stamp and that network's listings BEFORE canonicalizing, builds through a separate `buildTokenIndex` helper, and writes the cache only when `catalogUpdated` still equals the snapshot stamp; a refresh landing mid-build sends it around the loop to rebuild from the new catalog (three attempts, after which the quote is served from the current snapshot uncached rather than caching a possibly-behind index). No path can stamp an old index with a new stamp any more.

src/swap/central/stealthex.ts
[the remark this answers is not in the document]
j0ntz: Reversed in the latest fixup. The `rate-type-fallback-scope` invariant added to AGENTS.md and .cursor/BUGBOT.md on this branch scopes a fixed-then-float fallback to the failure it compensates for, and a limit rejection is not that failure: the route exists and the amount is what StealthEX refused, so retrying it on floating reports a different pair's limit than the one the quote is on. `checkLimits` throws again and `attemptRate` returns empty ONLY when the rate type has no route, with the comment now naming the missing provider capability (no per-route rate-type reporting) and the condition that would retire the fallback.

src/swap/central/stealthex.ts
```diff
@@ -0,0 +1,747 @@
...
+    swapInfo,
+
+    async fetchSwapQuote(req: EdgeSwapRequest): Promise<EdgeSwapQuote> {
+      const request = convertRequest(req)
+
+      checkInvalidTokenIds(INVALID_TOKEN_IDS, request, swapInfo)
+
+      const newRequest = await getMaxSwappable(getQuote, request)
+      const swapOrder = await getQuote(newRequest)
+      return await makeSwapPluginQuote(swapOrder)
```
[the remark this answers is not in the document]
j0ntz: Valid, and fixed. The quote path is split: `fetchQuote(request, enforceMax)` resolves the route, checks limits and returns both addresses without creating anything, `fetchProbeOrder` builds the probe `spendInfo` from it (own from-address, `skipChecks: true`), and `fetchSwapQuoteInner` is now the only caller that POSTs `/exchanges`. Above the maximum the probe clamps to the range max instead of skipping the check, because StealthEX rejects an out-of-range amount at `/rates/estimated-amount` with 400 InvalidAmount, which would have turned an above-limit max swap into an untyped error; the real quote still raises `SwapAboveLimitError`. Measured against the live API with the same harness on both sides of the change: a max quote issued 2 orders before and 1 after.

j0ntz:
## 📸 Test evidence: executed swap (TRX to ETH on Base) through the rebuilt plugin

**stealthex quote**

stealthex quote

**stealthex swap success**

stealthex swap success

j0ntz:
`AGENTS.md` was added in another PR that never merged. This PR rebased onto it on the latest round of changes and builds on top of it to further harden first-time agent swap implementations.

j0ntz:
### Adherence to #489

The plugin is written against `docs/API_REQUIREMENTS.md` as [#489](https://github.com/EdgeApp/edge-exchange-plugins/pull/489) revises it, and against the review rules in `.cursor/BUGBOT.md` that #489 adds alongside it.

Three of those rules came out of this integration and are in #489 rather than here:

- `per-endpoint-auth`: StealthEX authenticates every v4 endpoint including the catalog, so credentials are attached per request rather than assumed global.
- `catalog-cache-expiry`: the asset catalog is cached for an hour, replaced atomically, and the previous index is kept when a refresh fails. ARRR and ZANO were listed on 2026-08-14 and gone by 2026-08-18, so an unbounded cache would quote a delisted asset.
- `rate-type-fallback-scope`: the fixed-then-floating fallback fires only when the fixed route is missing. A limit error or an unsupported pair is not retried, and no order exists at that point, so no attempt can create a second one.

Where #489 is stricter than the doc this plugin was first written against, the plugin already complies:

| #489 requirement | Plugin |
|---|---|
| Native units govern every amount field; a display-unit API must document the convention | StealthEX quotes display units and documents nothing, so every amount is cleaned through a number-to-string cleaner before arithmetic, and `denominationToNative` converts at the boundary with integer rounding |
| Numeric EVM `chainId` (raised from **should** to **must**) | Not offered by StealthEX. The static chain table in `src/mappings/stealthex.ts` covers it, and the gap is written up rather than coded |
| Per-route rate types | Not offered. The scoped fallback above covers it |
| Max quote must not create an abandoned order | The max probe resolves route, limits and addresses without posting to `/v4/exchanges`; only the final quote creates the order |
| Provider amounts bounded before signing | A `from` quote rejects an order whose deposit amount exceeds what the user requested |

The compliance and gap report is regenerated against #489 and re-verified against the live API today: https://gist.github.com/j0ntz/ae71f231a5939489cf4611325058c00c

Section 2 (order identification) passes. Sections 1, 3, 4 and 5 are gaps on the StealthEX side, section 6 is partial, and section 7 needs a written answer from StealthEX rather than a code change. Issue 2 from the review is the section 1 gap.

j0ntz:
## 📸 Test evidence: account activation on an unactivated XRP wallet

**xrp reserve notice**

xrp reserve notice

**stealthex quote fresh xrp**

stealthex quote fresh xrp

**stealthex swap success fresh xrp**

stealthex swap success fresh xrp

**fresh xrp wallet funded**

fresh xrp wallet funded

**fresh hedera receive**

fresh hedera receive

j0ntz:
## 📸 Test evidence: funded activation on ALGO, XLM and SOL

**stealthex quote fresh algo**

stealthex quote fresh algo

**stealthex swap success fresh algo**

stealthex swap success fresh algo

**fresh algo wallet funded**

fresh algo wallet funded

**xlm min balance notice**

xlm min balance notice

**stealthex quote fresh xlm**

stealthex quote fresh xlm

**stealthex swap success fresh xlm**

stealthex swap success fresh xlm

**fresh xlm wallet funded**

fresh xlm wallet funded

**stealthex quote fresh sol**

stealthex quote fresh sol

**stealthex swap success fresh sol**

stealthex swap success fresh sol

**fresh sol wallet funded**

fresh sol wallet funded

**fresh hedera placeholder address**

fresh hedera placeholder address

j0ntz:
Review items, current state on `5c0a70d`.

| # | Item | State |
|---|---|---|
| 1 | must fix: token identity guessed from the Edge currency code | Fixed. `resolveListing` resolves a token through `wallet.currencyConfig.getTokenId` on the listing's `contract_address`, so the wallet's own currency plugin canonicalizes the address and a ticker never decides identity. |
| 2 | warning: EVM chains selected through a static network table, not the numeric chain ID | Not fixed, and not fixable against v4. See below. |
| 3 | must fix: resolved currencies cached for the plugin's lifetime | Fixed. The catalog carries a one hour TTL, and each derived token index is keyed on the catalog snapshot it was built from, so an index can never outlive its catalog. |
| 4 | must fix: every currency lookup failure reported as an unsupported pair | Fixed. A catalog outage now throws the provider error (or `SwapPermissionError` on `NotAllowed`) instead of claiming the pair does not exist. |
| 5 | must fix: `/rates/range` maps `NotAllowed` differently from the estimate and exchange paths | Fixed. All three paths go through one `stealthexError()`. |
| 6 | must fix: fixed-rate orders created without requiring `rate_id` | Fixed. A fixed estimate that carries no `rate.id` is treated as having no fixed route, so an order can never be created claiming a locked rate it does not hold. |

### Item 2

The v4 API this plugin targets exposes no chain ID. An asset is addressed only by `{symbol, network}`, where `network` is StealthEX's own name for the chain, so the table is the only way to name a chain.

StealthEX has since built an Edge-specific flow (`https://api.stealthex.io/docs/edge`) that does take `{chain_type: 'evm', chain_id, contract_address}`, which is exactly what this item asks for. It is not adoptable yet: probing all 48 mapped chains against it, 17 non-EVM chains have no network code at all and every candidate returns `404 NoPair` (monero, digibyte, cosmoshub, sui, filecoin, eos, fio, zano, zcoin, qtum, pivx, thorchainrune, wax, osmosis, ecash, bitcoinsv, axelar), and the flow publishes no catalog endpoint to look codes up with. Its `network` namespace also cannot address L2 natives at all: `network: "eth"` resolves to Ethereum mainnet for Base, Arbitrum, Optimism and zkSync alike. Moving now would drop those chains. The missing catalog is already open with StealthEX.

### Proactive fix in this push

Validating the mapping against the live catalog turned up a live defect this PR shipped with. The plugin resolved a chain's native asset as `symbol === mainnetSymbol && contract_address == null`, but StealthEX gives some native listings a pseudo-contract string rather than null: ETH on `base` is `baseeth`, AVAX on `avax-c` is `cchain`, ATOM is `atom1`, Vaulta is `core.vaulta`, and AXL is an IBC denom. Those five chains resolved to nothing, so StealthEX dropped out of the quote race for them with no error anywhere.

`{symbol, network}` is unique across all 1005 catalog listings, so the null-contract predicate bought nothing and the mapped pair alone identifies the asset. Pseudo-contracts are not addresses on their own chain, so they never survive into a token index and cannot collide with a real token.

Driving the real plugin code against the live API, before and after:

```
BEFORE
  UNRESOLVED base -> bitcoin: SwapCurrencyError
  UNRESOLVED avalanche -> bitcoin: SwapCurrencyError
  UNRESOLVED cosmoshub -> bitcoin: SwapCurrencyError
  UNRESOLVED eos -> bitcoin: SwapCurrencyError
  UNRESOLVED axelar -> bitcoin: SwapCurrencyError
  PASS monero -> bitcoin: quoted 621246

AFTER
  PASS base -> bitcoin: quoted 30636
  PASS avalanche -> bitcoin: quoted 19482
  PASS cosmoshub -> bitcoin: quoted 2321520
  SwapAboveLimitError eos -> bitcoin      (resolved; probe amount out of range)
  SwapBelowLimitError axelar -> bitcoin   (resolved; probe amount out of range)
  PASS monero -> bitcoin: quoted 621246   (control, unchanged)
```

Native chains resolving against the live catalog go from 39 to 44 of 48.

The same push adds a `log.warn` when a mapped chain is absent from a non-empty catalog. Four mapped chains have been delisted by StealthEX since this branch was written (bitcoinsv, piratechain, zano, fio) and were indistinguishable from an unsupported pair. They stay in the map so they re-list themselves if StealthEX brings them back.

j0ntz:
## 📸 Test evidence: funded StealthEX swap from Base ETH

**stealthex quote**

stealthex quote

**stealthex success**

stealthex success

PR #490, opened 2026-08-26
Require address-format handling from providers

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

Documentation only. No plugin code changes.

Three address-handling defects turned up while investigating Xgram and SimpleSwap against their live APIs ([Asana](https://app.asana.com/0/1215088146871429/1217836790411372)). Each was permitted by the spec as written, and each one reaches the user as a swap that cannot be quoted:

| Defect | Xgram | SimpleSwap |
|---|---|---|
| A standardized format rejected outright | bech32 Litecoin (`ltc1...`) refused, legacy and P2SH accepted | unified Zcash (`u1...`) refused, transparent and TEX accepted |
| A case variant treated as a different address | all-uppercase bech32 refused | `validationAddress` regexes match lowercase prefixes only |
| A length pinned into validation | n/a | `^(ltc1)[0-9A-Za-z]{39}$` excludes longer P2WSH and bech32m forms |

In both format-rejection cases the refused form is the one the Edge wallet hands out by default for that chain, so the pair fails for every user rather than some of them. Neither provider returns a distinct error code for it, so the rejection is indistinguishable from an unsupported pair and the user is told the pair is not supported.

This adds:

- **A new `Address formats` general principle**, split into two requirements of deliberately different weight. **Format coverage may be partial** and must be documented per chain: transparent, shielded and unified are different constructions, and a provider may reasonably support some and not others, so what the spec demands is that the gap is written down rather than discovered by a failed swap. **Encoding equivalence is not optional**: once a format is accepted, every valid encoding of an address in it is the same address, compared by decoding rather than by string equality, with no length or checksum constant pinned into validation. Case, optional prefixes, alternate account encodings and checksum revisions (bech32m for taproot) appear as examples rather than as the list.
- **Case-insensitive contract matching, in section 1.** Xgram resolves `0xdac17f9...` and refuses `0xdAC17F9...` for the same token, and its own asset list returns a mix of both casings. Section 1 also now requires the asset-list endpoint to name chains using the identifiers the quote endpoint expects. Xgram's differ (`Litecoin` in the list, `LTC` at the endpoint), which is why Litecoin and Stellar were never mapped on our side.
- **Two rows and a no-overloading rule, in section 3.** An unacceptable address and a retryable failure each need their own machine-readable code. Xgram currently returns `CURRENCY_UNSUPPORTED` for at least five distinct conditions, which is what collapses all of the above into "No enabled exchanges support X to Y".

Asset coverage stays a per-partner commercial matter and is deliberately not restated here. The Zcash unified-address requirement, for one, is already an Exhibit A term for SimpleSwap, and enumerating chains in this doc would imply the list is the requirement.

Section numbering is unchanged, so the `#3-error-handling` deep links already shared with partners still resolve.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - added: A test suite for the swap plugin template, covering the max-quote probe, limit handling, the trust boundary on provider amounts, and memo cleaning. A new integration can copy it alongside the template.
 - changed: The swap plugin template now models the max-quote probe, the trust boundary on provider-returned amounts, integer native-amount rounding, and funds-safe memo cleaning, so a new integration starts from them rather than rediscovering them in review.
 - changed: `docs/CREATING_AN_EXCHANGE_PLUGIN.md` covers max quotes and amount rounding in depth, and ends with a pre-PR checklist drawn from findings on shipped integrations.
+- changed: `docs/API_REQUIREMENTS.md` requires providers to document which address formats they accept per chain, to accept every encoding of a format they do support rather than string-matching what the caller sent, to match EVM contract addresses case-insensitively, to keep their asset list in agreement with what the order endpoint accepts, both in chain identifiers and in availability, and to give a rejected address and a retryable failure their own error codes.
 
 ## 2.53.0 (2026-08-18)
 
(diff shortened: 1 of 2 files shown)


j0ntz:
Force merging doc-only change

PR #491, opened 2026-08-28
Let LI.FI pick the slippage for each pair

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

Asana: https://app.asana.com/0/1215088146871429/1216079900947419

`lifi.ts` sent a hard-coded `slippage=0.05` on every `v1/quote` request, including stable-to-stable pairs. That 5% is the exact size of the window a sandwich bot can extract, and support has escalated four separate cases of users losing the whole of it (2760 USDT to USDC losing $160; 2.51 ETH to USDT losing about $650).

LI.FI computes its own slippage per pair when the parameter is omitted, so this drops the override. Measured against the live API, same request, only `slippage` differing:

| Pair | Edge's 5% | LI.FI's own |
|---|---|---|
| 2760 USDT to USDC (Ethereum) | min 2610.24 USDC | 0.1%, min 2744.87 USDC |
| 2.51 ETH to USDT (Ethereum) | min 5926.92 USDT | 0.5%, min 6207.67 USDT |
| 0.1 ETH (Ethereum) to POL | min 2187.53 POL | 0.5%, min 2290.86 POL |
| 1000 USDC Ethereum to Polygon | min 995.00 USDC | 0.1%, min 995.00 USDC |

So 0.1% for stable-to-stable and 0.5% otherwise. The plugin already surfaces `toAmountMin` as `minReceiveAmount`, so the "Min Receive Amount" the user sees becomes honest for free.

The plugin also reads an optional `swap.plugins.lifi.slippage` (a decimal, ie `0.01` for 1%) so a pair that turns out to need a looser band can be pinned without a client release. That override is inert until the info server serves it: `edge-info-server`'s own `asExchangeInfo` enumerates `swap.plugins` as `thorchain` / `thorchainda` / `swapkit` / `mayaprotocol` and drops unknown keys, so anything under `plugins.lifi` is stripped before it reaches us. Adding a `lifi` block there is a few lines, and it would also revive `lifiServers`, which this plugin has been reading and never receiving for the same reason.

Out of scope, worth a follow-up: `rango.ts` carries the same hard-coded 5%, and LI.FI's `preset=stablecoin` (their 1:1 stablecoin routes) is a real preset name but is rejected for our integrator until it is registered in the partner portal.

---

> [!NOTE]
> **Medium Risk**
> Changes swap quote and min-receive semantics for all LI.FI routes; tighter slippage may cause more quote failures on illiquid pairs until the optional info-server override is available.
> 
> **Overview**
> LI.FI **v1/quote** requests no longer send a hard-coded **5% slippage**. When no override is configured, the plugin omits the parameter so LI.FI applies its own pair-specific limits (typically **0.1%** stable-to-stable and **0.5%** otherwise), which tightens quoted **min receive** amounts and reduces sandwich-extraction exposure.
> 
> The plugin can still pass slippage when **`swap.plugins.lifi.slippage`** arrives from exchange info (decimal, e.g. `0.01` for 1%), wired through the existing **`asExchangeInfo`** schema alongside **`lifiServers`**. That override stays inactive until the info server exposes a **`lifi`** block; the CHANGELOG notes the same gap for **`lifiServers`**.
> 
> The removed **`MAX_SLIPPAGE`** constant and the conditional query-param spread avoid **`makeQueryParams`** emitting a bare **`slippage`** key for undefined values.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d95c9c76a29ad165c8205e1a1870b9be2551b14a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


j0ntz:
## 📸 Test evidence

**lifi quote min receive**

lifi quote min receive

**lifi swap success**

lifi swap success

**swap tx details**

swap tx details

peachbits (approved):

PR #493, opened 2026-09-03
Surface Xgram transport refusals

j0ntz:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

Xgram answers a rejected API key with an HTTP 200 whose body is
`{"status":401,"message":"Unauthorized"}`. That shape matches none of the quote
cleaners, so the plugin's cleaner union threw
`TypeError: Expected a string, got undefined at .error`, and the real cause never
reached the logs or the user. The plugin also retried the same refusal as a
float order, doubling the failed calls against an endpoint Xgram has separately
asked us to call less.

This adds a cleaner branch for that envelope, throws
`Xgram: HTTP <status> <message>` instead, and classifies it as non-retryable so
the float fallback stays scoped to the fixed-rate-only failures it exists for.
The rename from `isTypedSwapError` to `isNonRetryableError` reflects that the
guard now covers both.

Found while re-testing Xgram's API for
[the partner comms task](https://app.asana.com/0/1215088146871429/1218117826000804):
our partner key is currently rejected on every Xgram endpoint, so the app hits
this path on every Xgram quote today.

Tests: three unit cases cover the named error, the absent retry, and the float
fallback still firing on a rate-type failure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - added: Robinhood Chain swaps via SideShift, LI.FI, Rango, ChangeNow, Swapuz and LetsExchange.
+- fixed: Xgram now reports a rejected API key as `Xgram: HTTP 401 Unauthorized` instead of a cleaner type error, and stops retrying such a refusal as a float order.
 
 ## 2.54.0 (2026-08-28)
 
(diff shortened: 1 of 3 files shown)


j0ntz:
## 📸 Test evidence

**xgram quote refused**

xgram quote refused

